PR #133, opened 2015-08-08
wkhtmltopdf installation for Ubuntu 14.04 trusty

GhassenRjab:
Updated setup_frappe.sh to allow the installation of wkhtmltopdf for Ubuntu 14.04 trusty. In the past it led to 'is not a debian format archive' error. I also suggest to stop adding mariadb repositories for Ubuntu trusty as they are available by default.

### DIFF
--- a/install_scripts/setup_frappe.sh
+++ b/install_scripts/setup_frappe.sh
@@ -17,11 +17,11 @@ get_passwd() {
 
 set_opts () {
 	OPTS=`getopt -o v --long verbose,mysql-root-password:,frappe-user:,bench-branch:,setup-production,skip-setup-bench,help -n 'parse-options' -- "$@"`
-	 
+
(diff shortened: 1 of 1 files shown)


pdvyas:
Hey, I tested on a trusty install and `wkhtmltopdf` installs fine from the downloads at http://wkhtmltopdf.org/downloads.html

GhassenRjab:
It didn't work on my server until I made those changes. I will test it again.

anandpdoshi:
I installed on DO with 14.04 without any issues

GhassenRjab:
You are both right. I did the installation on two trusty machines (32 bits and 64 bits) and everything went fine. I still don't know what happened so that error came up to me in my previous installation which led me to make this pull request. Anyway, I just wanted to help. Thank you for the good project.

anandpdoshi:
Maybe that file for wkhtmltopdf wasn't available when you first tried it and it became available soon after or maybe there was a server issue.

Thanks for contributing 😀

PR #160, opened 2015-11-20
Revert "Automate swap partition setup for centos / ubuntu / debian servers"

anandpdoshi:
Reverts frappe/bench#159

@yashodhank Discussed with @pdvyas that changing swappiness is not a good idea. He will elaborate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,12 +26,6 @@ This script should install the pre-requisites, install bench and setup an ERPNex
 
 If you want to develop ERPNext or any Frappe App, you can omit the "--setup-production" part from the command.
 
-For advanced users, `setup_frappe.sh` script offers following options:
-```
-sudo bash setup_frappe.sh --setup-swap --setup-production
-```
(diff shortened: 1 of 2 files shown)


pdvyas:
I like the feature, just want make sure there are no edge cases. Have to be careful when changing kernel parameters and changing fstab. For example, it might try to remove a swap partition on disk (remove_old_swap..)

yashodhank:
@pdvyas Do you want me to drop remove_old_swap and swappiness from the script and send pull request again?

pdvyas:
yeah, apart from that I am not sure if setting it memory*3 and setting vfs_cache_pressure and swappiness (value other than 1) is a good idea. 

Presently, I am working on a port of this script to ansible and using this, https://github.com/pdvyas/deploy-erpnext/blob/master/ansible/roles/swap/tasks/main.yml (uses dd, fallocate is a better choice)

PR #171, opened 2015-12-25
rpm, deb, brew packages + ansible configuration

nginn:


pdvyas:
* We can't check in a  tar.gz everytime, please have a single step command/script that takes the current tree and makes the package.
* The ansible roles do not appear to be idempotent, refer, https://github.com/frappe/bench/tree/vm/vm/ansible
* The rpm does not work, it has only 4 files
```
# rpm -qpl bench-0.92-1.noarch.rpm
/usr/bin/bench
/usr/share/doc/bench-0.92
/usr/share/doc/bench-0.92/LICENSE.md
/usr/share/doc/bench-0.92/MANIFEST.in
/usr/share/doc/bench-0.92/README.md
```
* The deb doesn't put bench in the path, should be symlinked or something from `/usr/share/python/bench/bin/bench`
* Unable to build on test machine, any clue? http://dpaste.com/3FBZ6W3
* brew formula installs but does not work. Maybe pip installing the bench's requirements.txt into `/usr/local/Cellar/bench/0.92/libexec/lib/python2.7/site-packages/`is a better idea.
```
% bench
  File "/usr/local/Cellar/bench/0.92/libexec/bin/bench", line 5, in <module>
    from pkg_resources import load_entry_point
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/Extras/lib/python/pkg_resources.py", line 2793, in <module>
    working_set.require(__requires__)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/Extras/lib/python/pkg_resources.py", line 673, in require
    needed = self.resolve(parse_requirements(requirements))
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/Extras/lib/python/pkg_resources.py", line 576, in resolve
    raise DistributionNotFound(req)
pkg_resources.DistributionNotFound: honcho
```

nginn:
> Maybe pip installing the bench's requirements.txt is a better idea

No, it's not, see [here](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Python-for-Formula-Authors.md) <br>
Because we can't rely on pip in the brew formula I've come with the following solution: <br>
* all dependencies (including those which are automatically installed by pip) are listed explicitly as resources
* honcho is listed as formula dependency because it's distributed only as wheel module which cannot be installed by launching `python setup.py`
* after honcho installation, ruby code launches bash commands to find honcho installation path and to create symlinks into bench site-packages directory

pdvyas:
Yup, you're right :)
 Any solution for the honcho issue? Maybe you can try using this link,
https://github.com/nickstenning/honcho/archive/v0.6.6.tar.gz

nginn:
@pdvyas I've posted solution in the commit 2edeaa4 and described it in the previous comment, brew formula works ok now

pdvyas:
* brew install is stuck for at `writing list of installed files to 'installed.txt` infinetely for me (tried a bunch of times), are you able to reproduce this?
* deb works, one minor issue is it doesn't remove the `/usr/bin/bench` symlink on removing the package.
* rpm works, it does not appear that it uses the virtualenv method like deb as having python-requests installed did not allow me to install the package as the files were conflicting.
* rpm build failed on me with following, http://dpaste.com/2QWGWYM , any suggestions?
* The deb package breaks the system if deps not met, not sure if can be avoided.
http://dpaste.com/07CXVS7

nginn:
> brew install is stuck for at writing list of installed files to 'installed.txt infinetely for me (tried a bunch of
> times), are you able to reproduce this?

No, I'm not, using El Capitan version. On which version do you have this problem?

> deb works, one minor issue is it doesn't remove the /usr/bin/bench symlink on removing the package.

Fixed in 8596f93

> rpm works, it does not appear that it uses the virtualenv method like deb as having python-requests
> installed did not allow me to install the package as the files were conflicting.

yes, that's because dh-virtualenv doesn't work for rpm. I'll try to reproduce this confilct error

> The deb package breaks the system if deps not met, not sure if can be avoided. http://dpaste.com/07CXVS7 

That's why I have suggested usage of `gdebi` in the documentation, please have a look: https://github.com/nginn/bench/blob/master/installers/deb_package_build/readme.md

PR #173, opened 2016-01-08
VM build script

pdvyas:


vm/ansible/roles/mariadb/tasks/centos.yml~
```diff
@@ -0,0 +1,10 @@
+---
```
anandpdoshi: looks like a bad file with extension `yml~`?

vm/ansible/playbook.yml
```diff
@@ -0,0 +1,160 @@
+---
```
anandpdoshi: why 2 files? `base.yml` and `playbook.yml`? both look quite similar.

anandpdoshi:
@pdvyas :+1:

vm/ansible/playbook.yml
pdvyas: Gotta remove `playbook.yml`

PR #176, opened 2016-01-22
Enable auto update

revant:


revant:
Is auto update stable now?
It is documented in bench readme 
https://github.com/frappe/bench#updating

PR #183, opened 2016-02-22
[Fix] To disallow sym-link creation

shreyasp:
In case symlink for supervisor.conf and nginx.conf already exists in the bench/config folder, then by pass the creation of symlinks.


shreyasp:
@anandpdoshi Have a look at the fix, and let me know if I need to make a change in the fix.

PR #192, opened 2016-03-16
Added a new argument to setup_frappe.sh

omaraltayyan:
Add a new "--site" argument to the setup file, so that if someone wants to create their ERPNext setup with a DNS Multitenant setup, they can specify the new site's name before installation, without having to rename the site after the installation, or having to delete it and install a different one.

### DIFF
--- a/install_scripts/setup_frappe.sh
+++ b/install_scripts/setup_frappe.sh
@@ -16,7 +16,7 @@ get_passwd() {
 }
 
 set_opts () {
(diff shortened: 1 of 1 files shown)


anandpdoshi:
@vjFaLk can you incorporate this in the new installer? Ideally if site_name is not passed, the installer should ask for one, and should also ask if the user wants to setup dns mutlitenancy or wants port based multitenancy.

PR #198, opened 2016-04-02
Update README.md

ckosiegbu:
Added the use of the curl command on OS X as wget does not come by default on this platform.


anandpdoshi:
@vjFaLk can you manually incorporate what is required from this in the new readme? This is not mergable now. Make sure you attribute @ckosiegbu 

Essentially, we have to use `curl` instead of `wget` on Mac. Also see the other instructions in the commit. The 'super user' thing will be good to mention.

PR #205, opened 2016-04-21
[Fix] Fixed drop site test case

shreyasp:
- Fixed the test case for drop site.

### DIFF
--- a/bench/tests/test_init.py
+++ b/bench/tests/test_init.py
@@ -57,7 +57,7 @@ class TestBenchInit(unittest.TestCase):
 		})
 
 	def test_new_site(self):
-		self.test_init()
+		self.init_bench('test-bench')
 		self.new_site("test-site-1.dev")
 
 	def new_site(self, site_name):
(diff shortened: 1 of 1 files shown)


bench/tests/test_init.py
```diff
@@ -57,11 +57,10 @@ def test_multiple_benches(self):
 		})
 
 	def test_new_site(self):
+		self.init_bench('test-bench')
 		self.new_site("test-site-1.dev")
 
 	def new_site(self, site_name):
-		self.test_init()
```
anandpdoshi: Why remove this? The only problem was that `archived_sites_path` variable was missing

shreyasp: @anandpdoshi 

Previously each call to `new_site()` would call `test_init()` to create new bench and then create a new site. 

In case of `test_drop_site` I was trying to test both scenarios in a single test case(**with and without** ***archived_site_path***), and so `tearDown()` never happened which resulted into failure even after correcting the varibale name. And so, I moved call to `test_init()` to caller function, to avoid creation of bench for these two scenarios.

Also, I replaced it with `init_bench()` as we were executing `test_init()` for all the tests to create a new bench.

PR #210, opened 2016-04-29
Added validation for App Name

vjFaLk:
Fix for https://github.com/frappe/frappe/issues/1737

Sync with https://github.com/frappe/frappe/pull/1739

### DIFF
--- a/bench/app.py
+++ b/bench/app.py
@@ -62,7 +62,8 @@ def get_app(app, git_url, branch=None, bench='.', build_asset_files=True, verbos
 		restart_supervisor_processes(bench=bench)
 
 def new_app(app, bench='.'):
-	app = app.lower().replace(" ", "_")
+	# For backwards compatibility
+	app = app.lower().replace(" ", "_").replace("-", "_") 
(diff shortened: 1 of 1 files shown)


bench/app.py
```diff
@@ -62,7 +62,9 @@ def get_app(app, git_url, branch=None, bench='.', build_asset_files=True, verbos
 		restart_supervisor_processes(bench=bench)
 
 def new_app(app, bench='.'):
-	app = app.lower().replace(" ", "_")
+	app = app.lower().replace(" ", "_").replace("-", "_")
+	if not re.match("^[a-z_]*$", app):
```
anandpdoshi: This is validation done in frappe isn't it?

We should duplicate the lowercase and replace in frappe. It should've been in there from the start, but for backward compatibility, we will have to keep it here (put a comment about backward compat)

anandpdoshi:
Remove the regex validation from here and keep it only in frappe

anandpdoshi:
Add `app = app.lower().replace(" ", "_").replace("-", "_")` in frappe too

PR #213, opened 2016-04-30
make bench release available for app developers

revant:


bench/commands/utils.py
```diff
@@ -111,13 +111,14 @@ def backup_all_sites():
 
 
 @click.command('release')
-@click.argument('app', type=click.Choice(['frappe', 'erpnext', 'erpnext_shopify', 'paypal_integration', 'schools']))
+@click.argument('app')
 @click.argument('bump-type', type=click.Choice(['major', 'minor', 'patch']))
+@click.argument('owner', default='frappe')
```
anandpdoshi: The right term here should be `fork` instead of `owner`

i.e. `bench release erpnext patch --fork=revant`

anandpdoshi:
You will also need to modify `release.py` as it has a repo_map

bench/commands/utils.py
revant: I planned to use it as,
``bench release civil_contracting patch --(owner|fork)=revant``

revant:
What to add in ``repo_map`` in case of custom app?

anandpdoshi:
@revant you should remove the `repo_map` system itself. Try if you can do it, or I will take over if you can't.

revant:
I'll try for some time. If it takes too long you can take over.

Any suggestions how to go about it?

My first approach will be remove repo_map and trace/debug errors.

revant:
@anandpdoshi Releases seems to work on following command after https://github.com/frappe/bench/pull/213/commits/1102670b992f4a88c0509923015f3aaa42bfdcfd
```
bench release --owner revant release_test minor
```
released here : https://github.com/revant/release_test/releases

anandpdoshi:
@revant will cherry-pick and merge manually

bench/release.py
```diff
@@ -207,7 +199,7 @@ def create_github_release(owner, repo, tag_name, log, gh_username=None, gh_passw
 			raise Exception, "No credentials"
 		gh_username = github_username
 		gh_password = github_password
-	repo = repo_map[os.path.basename(repo)]
+	repo = os.path.basename(repo)
```
anandpdoshi: The only problem with this assumption is that the repo_name will be same as the folder name. This is problematic in cases where folder_name is say, `frappeclient`, but repo name is `frappe-client`. So, we should add an option for --repo_name for such cases.

bench/commands/utils.py
```diff
@@ -116,8 +116,9 @@ def backup_all_sites():
 @click.option('--develop', default='develop')
 @click.option('--master', default='master')
 @click.option('--owner', default='frappe')
-def release(app, bump_type, develop, master, owner):
+@click.option('--repo-name')
+def release(app, bump_type, develop, master, owner, repo_name):
 	"Release app (internal to the Frappe team)"
 	from bench.release import release
-	repo = os.path.join('apps', app)
+	repo = os.path.join('apps', repo_name) if repo_name else os.path.join('apps', app)
```
revant: @anandpdoshi is this right?
I tested 2 commands on https://github.com/revant/release_test app : 

1. ``bench release --owner revant --repo-name release_test release_test patch``
1. ``bench release --owner revant release_test patch``

revant:
https://github.com/frappe/bench/pull/213/commits/ea8d13442ef2adefc64cda681b1a9086178b7351

tested with : 
local app name release_smart
repo name release_delete

revant:
or should I write a function?
```
def get_repo_or_app(repo, app):
	if os.path.basename(repo) != app:
		return os.path.join('apps', app)
	else:
		return repo
```

bench/commands/utils.py
anandpdoshi: No. It should be used just before pushing the release!

Basically, the folder name can be different from the repo name, and when pushing the release, you have to use the github repo name and not the folder name if it is different. See the line where I commented.

anandpdoshi:
fixed manually and merged

anandpdoshi:
@revant See this commit: https://github.com/frappe/erpnext/commit/227b238fb18c6695f3281fcd9615f92773dc89fe

You will have to modify your app files similarly. New apps will have such setup via boilerplate.

PR #215, opened 2016-05-05
Removed get-app app-name argument

vjFaLk:
- [ ] Modify get-app command in installers

### DIFF
--- a/bench/app.py
+++ b/bench/app.py
@@ -42,19 +42,29 @@ def write_appstxt(apps, bench='.'):
 	with open(os.path.join(bench, 'sites', 'apps.txt'), 'w') as f:
 		return f.write('\n'.join(apps))
 
-def get_app(app, git_url, branch=None, bench='.', build_asset_files=True, verbose=False):
-	logger.info('getting app {}'.format(app))
+def get_app(git_url, branch=None, bench='.', build_asset_files=True, verbose=False):
+	#Gets repo name from URL
+	repo_name = git_url.rsplit('/', 1)[1].rsplit('.', 1)[0]
+	logger.info('getting app {}'.format(repo_name))
 	shallow_clone = '--depth 1' if check_git_for_shallow_clone() else ''
 	branch = '--branch {branch}'.format(branch=branch) if branch else ''
 
-	exec_cmd("git clone {git_url} {branch} {shallow_clone} --origin upstream {app}".format(
+	exec_cmd("git clone {git_url} {branch} {shallow_clone} --origin upstream".format(
 				git_url=git_url,
-				app=app,
 				shallow_clone=shallow_clone,
 				branch=branch),
 			cwd=os.path.join(bench, 'apps'))
-	print 'installing', app
-	install_app(app, bench=bench, verbose=verbose)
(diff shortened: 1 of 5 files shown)


bench/app.py
```diff
@@ -42,17 +42,31 @@ def write_appstxt(apps, bench='.'):
...
 				shallow_clone=shallow_clone,
 				branch=branch),
 			cwd=os.path.join(bench, 'apps'))
+
+	#Retrieves app name from setup.py
+	app_path = os.path.join(bench, 'apps') + "/" +repo_name + "/setup.py"
+	with open(app_path) as searchfile:
+		for line in searchfile:
+			if "name" in line: 
+				app = re.findall("'(.*?)'", line)[0]
```
anandpdoshi: wrong assumption: it may either be single quote or double quote

A simple solution to this is not to go line by line but to use this regex:

```
with open(app_path, 'rb') as f:
    app_name = re.search(r'name\s*=\s*[\'"](.*)[\'"]', f.read().decode('utf-8')).group(1)
if repo_name != app_name:
    apps_path = os.path.join(os.path.abspath(bench), 'apps')
    os.rename(os.path.join(apps_path, repo_name), os.path.join(apps_path, app_name))
```

bench/app.py
```diff
@@ -42,17 +42,31 @@ def write_appstxt(apps, bench='.'):
...
-	exec_cmd("git clone {git_url} {branch} {shallow_clone} --origin upstream {app}".format(
+	exec_cmd("git clone {git_url} {branch} {shallow_clone} --origin upstream".format(
 				git_url=git_url,
-				app=app,
 				shallow_clone=shallow_clone,
 				branch=branch),
 			cwd=os.path.join(bench, 'apps'))
+
+	#Retrieves app name from setup.py
+	app_path = os.path.join(bench, 'apps') + "/" +repo_name + "/setup.py"
```
anandpdoshi: Poor way to write!

`setup_path = os.path.join(bench, 'apps', repo_name, 'setup.py')`

bench/commands/make.py
```diff
@@ -19,13 +19,12 @@ def init(path, apps_path, frappe_path, frappe_branch, no_procfile, no_backups,
 
 
 @click.command('get-app')
-@click.argument('name')
 @click.argument('git-url')
 @click.option('--branch', default=None, help="branch to checkout")
-def get_app(name, git_url, branch):
+def get_app(git_url, branch):
```
anandpdoshi: This will break things. So accept the `name` variable for backward compatibility as we discussed. Just add a comment about it.

Or, we will have to fix documentation, setup script, our deployment system, etc.

anandpdoshi:
@vjFaLk are you going to modify get-app commands or should I merge this?

anandpdoshi:
Also:

- [ ] Test Case: Try doing get-app for https://github.com/frappe/frappe-client and see if the foldername is set as `frappeclient`

vjFaLk:
@anandpdoshi  I've made changes within bench itself, will co-ordinate with @shreyasp and remove it from everywhere.

PR #227, opened 2016-05-24
[Fix] Added option to force upgrade nginx config file

shreyasp:
- While regenerating nginx configuration using `bench setup nginx` user is presented with a prompt. Adding `--force` to the command line will bypass this prompt.

### DIFF
--- a/bench/commands/setup.py
+++ b/bench/commands/setup.py
@@ -15,10 +15,11 @@ def setup_sudoers(user):
 
 
 @click.command('nginx')
-def setup_nginx():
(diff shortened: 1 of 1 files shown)


bench/commands/setup.py
```diff
@@ -15,10 +15,11 @@ def setup_sudoers(user):
 
 
 @click.command('nginx')
-def setup_nginx():
+@click.option('--force', help='Force regeneration of nginx config file', default=False, is_flag=True)
+def setup_nginx(force=None):
```
anandpdoshi: Also add this for supervisor and redis configs

PR #235, opened 2016-05-27
use wget instead of curl for ubuntu/debian

kressi:
correction of #228; because `curl` is not installed on a clean debian installation, nodesource script could not be downloaded with `curl`. therefore `npm` was not availabele after `apt-get install nodejs`. Download `nodesource` with `wget` instead.


anandpdoshi:
@vjFaLk / @shreyasp  can you do this in the new installer?

anandpdoshi:
@kressi please use the new installer instead

PR #236, opened 2016-05-30
Issues with Mac development bench installer

shreyasp:
- Setuptools to be installed using bootstrap script instead of `pip`
- Do not get passwords while running script to setup development environment
- Added option to mention `repo-url` along with the branch
- `extra_vars.json` file moved to `/tmp` directory instead of user home directory
- create user ansible script needs to perform `chown` only in case of Linux.


fellipeh:
Hello,

These PR, fiz these problem: https://github.com/frappe/bench/issues/237  ?

I'm trying to install and getting these error.

PR #246, opened 2016-06-24
mariadb repo check

slushpuppy:
Without mariadb repo check, setup script will fail for jessie 8.5 or any debian distro that has existing mariadb


rmehta:
@slushpuppy FYI, we will be soon moving to the new ansible based setup script.

PR #249, opened 2016-06-29
[Minor] Added command to archive-bench

shreyasp:
- [x] New command to archive bench `bench archive-bench {bench_path}`.
- [x] Delete symlinks from nginx and supervisor configs before archiving the bench in case of production.
- [x] Archived bench can be moved to default path or custom path provided by the user `--archived-benches-path`
- [x] Force archival in case site exists on the bench using `--force`.
- [x] Restarting and reloading `nginx` and `supervisorctl` after deleting symlinks.

### DIFF
--- a/bench/utils.py
+++ b/bench/utils.py
@@ -8,10 +8,11 @@ import json
 import platform
 import select
 import multiprocessing
-from distutils.spawn import find_executable
 import pwd, grp
 import bench
+import shutil
 from bench import env
+from distutils.spawn import find_executable
 
 class PatchError(Exception):
 	pass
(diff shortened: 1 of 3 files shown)


bench/utils.py
```diff
@@ -660,3 +661,85 @@ def validate_pillow_dependencies(bench_path, requirements):
...
+
+	if not archived_benches_path:
+		archived_benches_path = os.path.join(base_path, 'archived_benches')
+
+	if not os.path.exists(archived_benches_path):
+		os.makedirs(archived_benches_path)
+
+	shutil.move(bench_path, archived_benches_path)
+
+def restart_nginx_process():
```
shreyasp: 👎 Trivial code.. need to find out better way out to do this

PR #256, opened 2016-07-07
[Minor] Added 502 page to Nginx config

shreyasp:


anandpdoshi:
fixed and merged manually

PR #287, opened 2016-08-03
[Fix] Add 'allow_rate_limiting' as common_site_config to generate nginx config as per requirement

shreyasp:


rmehta:
related central fix?

Issue #257, opened 2016-07-11
bench restart web


rmehta:
command for restarting only web workers

will be great for hotpatch

PR #297, opened 2016-08-10
[Minor] Added command to restart only web workers in supervisor 

shreyasp:
- Use `bench restart --web-workers` to only restart web workers.


shreyasp:
Fixed #257

PR #306, opened 2016-08-22
update inputrc for history

vjFaLk:


playbooks/production/includes/setup_prod_env.retry
```diff
@@ -0,0 +1 @@
+localhost
```
rmehta: needed?

vjFaLk: Fixed

PR #309, opened 2016-08-28
Add git utility commands

phamviet:
New commands added:

1. bench remote-set-url `git_url`: Set upstream remote url for app
2. bench remote-reset-url:  Reset upstream remote url to frappe official
3. bench remote-urls: List apps remote url


vjFaLk:
@phamviet Hey, you need to make a PR to the master branch, the develop branch isn't being worked on anymore. 

Also, I'm not sure of the usefulness of this feature. There are git commands to do exactly those things, what's your use case?

phamviet:
Thanks @vjFaLk for your suggestion. Yes, there are similar git commands but i thought it could be useful to do all thing from a single place where bench tracks everything about installed apps. In my use case, i need a way to quickly switch to my forked repository on production server, that way we can ensure it stability.

PR #311, opened 2016-09-13
Fix develop installation on ubuntu 14.04 to 16.x

aarondiazr:
added two packages to complete installation of frappe

### DIFF
--- a/playbooks/install.py
+++ b/playbooks/install.py
@@ -14,7 +14,7 @@ def install_bench(args):
 	success = run_os_command({
 		'apt-get': [
 			'sudo apt-get update',
-			'sudo apt-get install -y git build-essential python-setuptools python-dev libffi-dev libssl-dev'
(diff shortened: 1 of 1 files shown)


playbooks/install.py
```diff
@@ -14,7 +14,7 @@ def install_bench(args):
 	success = run_os_command({
 		'apt-get': [
 			'sudo apt-get update',
-			'sudo apt-get install -y git build-essential python-setuptools python-dev libffi-dev libssl-dev'
+			'sudo apt-get install -y git build-essential python-setuptools python-dev libffi-dev libssl-dev libldap2-dev libsasl2-dev'
```
shreyasp: You should move these changes to ubuntu.yml as these are needed only for installation of frappe while all other required as dependencies to install Ansible, and run the script.

shreyasp: Also, Please rebase your branch with `Master` node installations have changed because of which the Travis is failing for your PR

PR #319, opened 2016-09-27
Move Easy Installer above Manual Install

vjFaLk:
It's better for non-advanced users

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,60 @@ If you have questions, please ask them on the [forum](https://discuss.erpnext.co
 
 ## Installation
 
+## Easy Install
+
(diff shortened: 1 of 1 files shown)


rmehta:
@vjFaLk this is by design. users must understand what the easy install does!

vjFaLk:
@rmehta Hmm, a user's first instinct would be to start with the top, maybe then add a link to the Easy Install at the top?

rmehta:
I think this reads alright the way it is... closing this.

vjFaLk:
💔

rmehta:
Sometimes its better to know what you are getting into :)

Issue #323, opened 2016-10-06
DNS Multinenant by default


rmehta:
Please do this on priority

vjFaLk:
I remember doing this, I think it was undone by Anand later for some reason? Anyway, this should fix this : 

https://github.com/frappe/bench/pull/324

PR #324, opened 2016-10-06
DNS Multitenancy by default

vjFaLk:
Fix for : https://github.com/frappe/bench/issues/323


rmehta:
Thanks @vjFaLk will fix it as soon as the tests pass!

PR #326, opened 2016-10-06
Added command to setup firewall

vjFaLk:
## `bench setup firewall`
Added `run_playbook` method to utils.py. Method exists in install.py, however, would have to make too many workarounds to re-use it. Currently dropped setup_firewall.yml inside `playbooks/production`, not sure of the best place. Command also checks if Ansible is installed.

### Note : firewalld doesn't work on Ubuntu 16.04 (with ansible) and UFW doesn't work on CentOS 7 😠 . Therefore the playbook now installs firewalld on CentOS and UFW on Ubuntu / Debian. Sure, it looks ugly, but it won't break. 🙈 

- [x] Testing: 
  - [x] Ubuntu
  - [x] Debian
  - [x] CentOS
- [x] Check if Ansible is installed, and promt to install if not

### DIFF
--- a/bench/cli.py
+++ b/bench/cli.py
@@ -47,7 +47,7 @@ def check_uid():
 		sys.exit(1)
 
 def cmd_requires_root():
-	if len(sys.argv) > 2 and sys.argv[2] in ('production', 'sudoers', 'lets-encrypt', 'fonts', 'reload-nginx'):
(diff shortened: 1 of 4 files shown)


shreyasp:
@vjFaLk This might not install the firewall for older systems as they won't have Ansible on their servers. We will need to make it backward compatible as well, so there might be some additional work needed to put into this PR.

vjFaLk:
@shreyasp Made fixes and tested on platforms.

Issue #232, opened 2016-05-26
Set InnoDB Buffer Pool in Production Setup


anandpdoshi:
Set in frappe.cnf, under `[mysqld]`
`innodb_buffer_pool_size=50% of memory`

PR #331, opened 2016-10-19
Set innodb buffer pool size in production setup

vjFaLk:
Fix for : https://github.com/frappe/bench/issues/232

Sets the InnoDB Buffer Pool Size to 50% of the total RAM during Production Setup


vjFaLk:
@shreyasp Accidental, shouldn't matter I believe

vjFaLk:
@shreyasp It seems Github / Travis has bugged out and is showing the tests aren't complete even though they passed successfully

PR #332, opened 2016-10-21
add symlink to setup nginx

RobertSchouten:


vjFaLk:
What's the idea behind this? Bench already creates a nginx.conf in the config folder that has a symlink to a file in conf.d

RobertSchouten:
The goal behind this is starting from a develop environment running Bench setup nginx seems confusing that it doesn't actually fully setup nginx

vjFaLk:
The development environment doesn't use nginx **at all**. What are you trying to achieve exactly?

shreyasp:
@RobertSchouten But in case of, development we serve sites using Werkzeug server which is started by command `bench serve --port <port>` in Procfile. We haven't used Nginx for serving our development sites.

RobertSchouten:
the case is using a develop vm in bridged mode so it has it's own ip and me being lazy and not wanting to type the port in to access the site

vjFaLk:
@RobertSchouten 😅 You can already do that if you have DNS Multitenant enabled, and no default site (no currentsite.txt in sites folder) just add entries to the /etc/hosts file for the domain name that point to the IP of your VM. I've been using that kind of setup for my own workflow. Let me know if you have any trouble with that on Discuss. 

I'm gonna close this PR 😄

RobertSchouten:
np just seemed like could of improved but havent gotten into dns multiteant but was in my todo so ill do that

PR #353, opened 2017-01-07
Fix syntax error in setup_frappe.sh

raphaelm:
This caused trouble for some other people too, see: https://discuss.erpnext.com/t/erpnext-debian-8-install-error/13988


vjFaLk:
Old installer is deprecated.

PR #365, opened 2017-03-07
Fix determine current app branch

wshurafa:
This fix to handle branches with slashes, when using gitflow for example, you will have branches with names hotfix/branch-name. The bench commands fails with these branches names because it ignores anything before the slash. This fix would fix this issue to get the full name of the branch name


revant:
For now we do not have hotfix branch for bench.
@wshurafa Can you send this PR into bench/develop

wshurafa:
https://github.com/frappe/bench/pull/465

Issue #421, opened 2017-05-30
--site is never used when installing with --develop


RogueRonin:
Documented nowhere that it isn't. 

I see no reason why it shouldn't setup a new site.

saurabh6790:
Fixed!

PR #423, opened 2017-05-31
--site is never used when installing with --develop #421

RogueRonin:
--site is never used when installing with --develop https://github.com/frappe/bench/issues/421


saurabh6790:
Added via https://github.com/frappe/bench/pull/473

PR #440, opened 2017-06-30
Added command to give git status for all apps in bench

vjFaLk:
Current command : `bench status`

Example Output : 

```
FRAPPE
INFO:bench.utils:git status -sb
## customv7...upstream/customv7
 M frappe/utils/oauth.py


ERPNEXT
INFO:bench.utils:git status -sb
## customv7...upstream/customv7
```


vjFaLk:
Is there a away to remove the INFO line?

PR #447, opened 2017-07-14
Create two virtualenvs for bench

tundebabzy:


tundebabzy:
I'll complete the task with multiple small commits

PR #460, opened 2017-07-27
Added docker specific changes to bench

vishalseshagiri:
- Added skip-bench-mkdir as a parameter to the bench init command which skips frappe_bench mkdir if it already exists. 
- Added skip-redis-config-generation as a click option to the bench init command to skip the redis.generate_config method
- Added a try catch block to catch and except file exists Error 17 exception [see lines 56-61 in utils.py]


vishalseshagiri:
Just found a bug, please dont merge this, will fix it.

PR #465, opened 2017-08-08
Fix determine current branch

wshurafa:
This fix to handle branches with slashes, when using gitflow for example, you will have branches with names hotfix/branch-name. The bench commands fails with these branches names because it ignores anything before the slash. This fix would fix this issue to get the full name of the branch name


wshurafa:
@revant

PR #473, opened 2017-08-20
Add default ERPNext installation to Easy Install script (develop)

rish4bhn:
@rmehta, I'm creating a site called **site1.default** for default ERPNext installation. Is this approach okay or should we name the site in another way?

Works great on Ubuntu 16.04, I will now test on all other OSes.

- [ ] Test on OS X
- [x] Test on Debian 7.1
- [x] Test on Centos 7.3
- [x] Test on Ubuntu 16.04
- [x] Fix incomplete README.md

### DIFF
--- a/playbooks/develop/centos.yml
+++ b/playbooks/develop/centos.yml
@@ -85,5 +85,5 @@
   - include: includes/setup_dev_env.yml
     when: not run_travis and not production
   
-  # setup an erpnext site called site1.default
+  # setup an erpnext site called site1.local
   - include: includes/setup_erpnext.yml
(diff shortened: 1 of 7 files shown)


rmehta:
`site1.local` would be a better name!

rish4bhn:
`site1.local` it is!

playbooks/develop/centos.yml
```diff
@@ -84,3 +84,7 @@
   # setup development environment
   - include: includes/setup_dev_env.yml
     when: not run_travis and not production
+  
+  # setup an erpnext site called site1.local
+  - include: includes/setup_erpnext.yml
```
rmehta: why not just add it to `setup_dev_env.yml`?

rish4bhn: Yes that makes more sense, I'll do it, thanks

PR #506, opened 2017-11-22
[fix] provision to setup maxretry, bantime and findtime explicitly for fail2ban

saurabh6790:


joolsr:
I have setup fail2ban on my server running ERPNext previously. Does this addition mean that fail2ban is setup fully and I should remove my config for fail2ban to avoid checking everything twice ? Are there any docs for this new addition to Bench? Great to see this !

PR #513, opened 2017-11-30
Install debian 9 support

jwrober:
All of the *.yml Ansible files already have support for Debian 9. The only thing missing is an update to Ansible 2.4.1 and then changing the test up front. This has been tested on a out of box install on Debian 9.2.1.

### DIFF
--- a/playbooks/install.py
+++ b/playbooks/install.py
@@ -115,7 +115,7 @@ def install_bench(args):
 		shutil.rmtree(tmp_bench_repo)
 
 def check_distribution_compatibility():
-	supported_dists = {'ubuntu': [14, 15, 16], 'debian': [7, 8],
+	supported_dists = {'ubuntu': [14, 15, 16], 'debian': [7, 8, 9],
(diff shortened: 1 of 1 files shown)


saurabh6790:
@jwrober Thanks!

codingCoffee:
@jwrober no additions need to be made for Debian 9 :)
But the script works due to backward compatibility but `include` has been deprecated in Ansible 2.4 and needs to be removed:

```
[DEPRECATION WARNING]: 'include' for playbook includes. You should use 'import_playbook' 
instead. This feature will be removed in version 2.8. Deprecation warnings can be disabled by
 setting deprecation_warnings=False in ansible.cfg.
[DEPRECATION WARNING]: The use of 'include' for tasks has been deprecated. Use 'import_tasks'
 for static inclusions or 'include_tasks' for dynamic inclusions. This feature will be 
removed in a future release. Deprecation warnings can be disabled by setting 
deprecation_warnings=False in ansible.cfg.
```

jwrober:
Ok, that is fine. Good to know on the depr warning. I'll do another PR with 2.4.1 and the fixes...

PR #524, opened 2017-12-28
Update README.md

ashish-greycube:
sudo python install.py --develop
This is NOT just develop mode of setup but uses branch develop. I have made that explicit


rmehta:
This is a non standard way, better to add a section for branches

PR #531, opened 2018-01-09
mariadb playbook fix for ubuntu and debian

codingCoffee:
- fixes travis builds


rmehta:
we don't need mariadb client anymore since we use pymysql ?

codingCoffee:
@rmehta its installed by default anyways, I just added it as a fallback

Issue #537, opened 2018-01-12
Fixes to bench setup lets-encrypt command


revant:
Related posts:

https://community.letsencrypt.org/t/2018-01-09-issue-with-tls-sni-01-and-shared-hosting-infrastructure/49996

https://community.letsencrypt.org/t/help-test-certbot-apache-and-nginx-fixes-for-tls-sni-01-outage/50207

Steps done manually (must be part of command `bench setup lets-encrypt`)

- Added lines to nginx.conf under site's server block listening to port 80

```
location '/.well-known/acme-challenge'
{
        default_type "text/plain";
        root /tmp/acme-challenge-public/;
}
```

example in frappe's context:

```
# http to https redirect
        server {
            listen 80;
                server_name
                        <sitename>
                        ;

            #return 301 https://$host$request_uri; # commenting redirect

            ### new lines ###
            location '/.well-known/acme-challenge'
            {
                    default_type "text/plain";
                    root /tmp/acme-challenge-public/;
            }
            ### new lines ###
       }
```

- reload nginx
- Add directory `/tmp/acme-challenge-public` and set owner `sudo chown -R root:root /tmp/acme-challenge-public`

- run certbot

Use this command for testing on staging server for development / testing
https://letsencrypt.org/docs/rate-limits/

```
sudo /opt/certbot-auto --verbose --staging --break-my-certs certonly -d <sitename> --webroot -w /tmp/acme-challenge-public/
```

for production

```
sudo /opt/certbot-auto --verbose certonly -d <sitename> --webroot -w /tmp/acme-challenge-public/
```

revant:
https://community.letsencrypt.org/t/tls-sni-challenges-disabled-for-most-new-issuance/50316

revant:
Easier workaround shared on forum

https://discuss.erpnext.com/t/latest-let-encrypt-error-january-10-2018-02-16-utc-investigating-the-tls-sni-challenge-has-been-disabled-due-to-strong-credibility-of-a-vulnerability-report/32524/10

jodeq:
I have test a modification in this PR #541 
It works in my instances and you could apply it easily

You must remove /etc/letsencrypt/configs/{site}.cfg if it exists because the last line should be `standalone-supported-challenges = tls-sni-01`

Then a `sudo bench setup lets-encrypt  {site}` is working

PR #541, opened 2018-01-16
Remove deprecated standalone-supported-challenges

jodeq:
With this modification I can now use bench setup lets-encrypt without any error.


saurabh6790:
@jodeq thanks for the PR!

revant:
fixes https://github.com/frappe/bench/issues/537

PR #542, opened 2018-01-17
fix-mysql-error

elba7r:
this pull request fix the mysql installation error


elba7r:
hello everyone i have found that there is an installation error appear and after a lot of investigations i found the reason which had been fixed in this pull request so please apply it because its a global error facing everyone start to install frappe since 8 days ago . thx

codingCoffee:
Hey @elba7r ,
this PR can be closed. I sent in a PR #543 targeting the root cause of the problem

yanncopp:
The Frappe Bench easy installation doesn't work !!!! Any solution ?????

saurabh6790:
already fixed via PR #543

PR #543, opened 2018-01-17
Fixed playbook to be compatible with Mariadb 10.2

codingCoffee:
- linked mysql.sock file mysql.pid file for compatibility with MaraiDB 10.2


jwrober:
This commit is not a good PR. Creating symlinks to `/var/run` from `/var/lib` for MariaDB pid and sock files violates basic Linux FHS standards! It would be much better to use the workaround that i have in my install steps documented here - <https://jwrober.github.io/erpnext_admin_guide/i-u-b/install>. Also the extra commands to "fix user mismatch" breaks the install completely. This PR needs to be reverted as soon as possible.

PR #544, opened 2018-01-18
fix for ubuntu and centOS

codingCoffee:


codingCoffee:
Don't merge

codingCoffee:
Merge-able :+1: 
tried and tested on Ubuntu 16.04 and CentOS7

@saurabh6790 please review
- Linking of mysql.sock and mysql.pid files was done for MariaDB 10.2 compatibility
- ansible_user_id used to give 'root' if the easy isntall script was executed by root on centOS, but it used to give 'frappe' if on Ubuntu, hence changed ansible_user_id to frappe_user

jwrober:
This change will break the usage of the --user flag to install.py just like https://github.com/frappe/bench/pull/543

jwrober:
@codingCoffee - your fix for the .cnf won't work. the `/etc/mysql/mariadb.conf.d` directory is read too early. This is why in my admin guide (https://jwrober.github.io/erpnext_admin_guide/i-u-b/install) I purposely put it in `/etc/mysql/conf.d/` as `z-erpnext.cnf`.

codingCoffee:
Hey @jwrober, first of all thanks for mentioning about the Linux FHS stadards. I didn't know about it and hence just tried to fix the problem by linking the file. 

Secondly, I also modified the `settings.cnf` as well, so as to include the `mariadb.conf.d` directory. Therefore it'll be read at the end of the `settings.cnf` as well, thereby fixing the problem of an early read.

codingCoffee:
Also can you comment on this:

> Also the extra commands to "fix user mismatch" breaks the install if you use the --user argument.

I tried executing the install script using the `--user` argument multiple times, it works perfectly

jwrober:
@codingCoffee - I looked over the changes and see where you put extra commands in settings.cnf to read the mariadb.conf.d stuff, so that should work. WRT to the --user flag issue, you added a collection of `become` and `become_user` directives to some of the files and these seems to have broken the flag. I see now you have removed some you added and added in other places. I am no expert w/ ansible scripts so I am unsure what those commands to. If there is a way to use these and fix the CentOs `/home/root` issue that I mention in the admin guide that is great. but the way those directives were in there yesterday when I tested and made my post it was totally broken if you used the command line argument during install and kind of worked when you did not use the argument as all files were placed in `/home/frappe`, which is not what I was expecting. i'll run another test later today and reply here.

PR #546, opened 2018-01-20
Bench setup failed on production fix

codingCoffee:
- mariadb 10.2 fix
- root user used instead of frappe_user fix
- selinux is set to permissive, in case of CentOS. This is needed to setup nginx
- bench is installed in the /home/username directory and not /home/root


codingCoffee:
Don't merge

PR #570, opened 2018-02-09
[fix] update setuptools and pip before creating virtual env and use system default python

saurabh6790:


achillesrasquinha:
All good!

PR #578, opened 2018-02-21
Prerequisites for EasyInstall and Cleanup

dottenbr:
I noticed that the instructions in EasyInstall were not so easy for novices and separated the prerequisites from the introduction text. It is clear now what exactly to do.


codingCoffee:
Added the required wiki links to README. Feel free to open if necessary.

Issue #575, opened 2018-02-20
EasyInstall broken


dottenbr:
**EayyInstall currently doesn't work.** 
Tested on DigitalOcean: 1 GB Memory / 25 GB Disk / LON1 - Ubuntu 16.04.3 x64

For details see latest post below.

skanel:
I have succeed for this installation.
1. make sure you run this `sudo python install.py --develop --user frappe` with other ubuntu user not a `frappe` user
2. your `frappe`user is normal user not administrator user 

Hope this help.

dottenbr:
I can not confirm this works. As root on a digitalocean droplet:

### Commands:

sudo apt-get update -y
sudo apt-get upgrade -y
apt-get install python-minimal -y
apt-get install build-essential python-setuptools -y

python install.py --develop --user frappe

### Output:

```
TASK [init bench] ***********************************************************************************
[...]

PR #580, opened 2018-02-21
installs yarn globally

codingCoffee:


codingCoffee:
probable fix for #575, dont merge, testing it out

PR #602, opened 2018-03-22
Allow dry-runs and non-interactivity 

vjFaLk:
This is to be able to automate release commands and also be able to compare branches


codingCoffee:
Closing, as I don't think we should have flags to bypass important stuff!

Issue #610, opened 2018-03-25
install.py fails because npm dependency unmet


oxwivi:
```
TASK [nodejs : Install yarn] ***************************************************
fatal: [localhost]: FAILED! => {"changed": false, "cmd": "npm install -g yarn", "failed": true, "msg": "[Errno 2] No such file or directory", "rc": 2}

RUNNING HANDLER [mariadb : restart mysql] **************************************
	to retry, use: --limit @/tmp/.bench/playbooks/site.retry

PLAY RECAP *********************************************************************
localhost                  : ok=48   changed=33   unreachable=0    failed=1   

Traceback (most recent call last):
  File "install.py", line 387, in <module>
    install_bench(args)
  File "install.py", line 109, in install_bench
    run_playbook('site.yml', sudo=True, extra_vars=extra_vars)
  File "install.py", line 325, in run_playbook
    success = subprocess.check_call(args, cwd=os.path.join(cwd, 'playbooks'))
  File "/usr/lib/python2.7/subprocess.py", line 541, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['ansible-playbook', '-c', 'local', 'site.yml', '-e', '@/tmp/extra_vars.json', '--become', '--become-user=frappe']' returned non-zero exit status 2
```
It should be one of the dependencies automatically installed by the script before this step.

oxwivi:
Okay, `nodejs` should've installed `npm`, but this is what the script outputs before the `npm` step:
```
TASK [nodejs : Add apt key for node repo] **************************************
changed: [localhost]

TASK [nodejs : Add repo] *******************************************************
changed: [localhost]

TASK [nodejs : Install nodejs 6.x] *********************************************
changed: [localhost]

TASK [nodejs : Install yarn] ***************************************************
fatal: [localhost]: FAILED! => {"changed": false, "cmd": "npm install -g yarn", "failed": true, "msg": "[Errno 2] No such file or directory", "rc": 2}
```
After facing this error and manually installing `npm`, the script outputs:
```
TASK [nodejs : Add apt key for node repo] **************************************
ok: [localhost]

TASK [nodejs : Add repo] *******************************************************
ok: [localhost]

TASK [nodejs : Install nodejs 6.x] *********************************************
ok: [localhost]
```
Maybe it didn't do `apt update` before installing `nodejs`?

oxwivi:
[...]

PR #615, opened 2018-03-26
Update main.yml

oxwivi:
Add support for Raspberry Pi and other ARM boards. I've been [failing](https://github.com/frappe/bench/issues/610) to install `bench` and subsequently, ERPNext for the sole reason of `node 6.x` not being downloaded and installed because the `apt` sources list only mentioned `amd64` and `i386` archs.


oxwivi:
I understand but manually specifying `arch` is pretty useless. Like the MariaDB's repository add that doesn't specify any, system automatically searches for the correct arch.

Basically, you can just leave it blank.

PR #624, opened 2018-04-05
bench setup systemd

revant:
To Do : 

- [x] command to add symlinks and `systemctl daemon-reload`
- [x] command to remove symlinks `systemctl daemon-reload`
- [x] background workers (numprocs substitute) https://unix.stackexchange.com/a/288310


revant:
bench_systemd

PR #642, opened 2018-04-17
Pip10 compatible 

codingCoffee:


codingCoffee:
fixes #690

dottenbr:
We are now at PIP 18.0, any reason this one should not be merged?

PR #643, opened 2018-04-17
Fix installation error - Mac

msabukwaik:
Fix : 
matplotlib 1.3.1 requires nose, which is not installed.
matplotlib 1.3.1 requires tornado, which is not installed
screen shot 2018-04-17 at 5 23 13 pm

After editing requirements.txt, the error has been resolved 
screen shot 2018-04-17 at 5 23 36 pm

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,6 @@
 nose 
 tornado
+matplotlib
 Click
 jinja2
 virtualenv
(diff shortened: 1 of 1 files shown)


codingCoffee:
Aaand why do we need matplotlib again ? 
You have added matplotlib as a dependency, and then added it's dependencies (nose and tornado) for no reason.  
Closing the PR. Feel free to open again if need be :wink:

PR #647, opened 2018-04-24
use `witchcraft.which to get the correct python executable

tundebabzy:
I discovered that bench was always making use of `/usr/bin/python` despite the fact that I was using an activated virtualenv. It turns out that bench depends on `find_executable` to get the python executable which isn't a reliable way. Python 3 has `shutil.which` which behaves exactly like you are running `which <something>` on the command line. The problem is it is not backported to python 2.7 so I added a dependency (whichcraft) which backports it to python 2.

I don't know how this can be tested but to reproduce it, activate a virtualenv and run `bench init`. You'll discover that your virtualenv's python executable won't be used

### DIFF
--- a/bench/utils.py
+++ b/bench/utils.py
@@ -1,9 +1,23 @@
-import os, sys, shutil, subprocess, logging, itertools, requests, json, platform, select, pwd, grp, multiprocessing, hashlib
+import grp
+import itertools
+import json
+import logging
+import multiprocessing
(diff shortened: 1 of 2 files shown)


codingCoffee:
I think it distutils works perfectly in python 2 and 3 alike

screenshot_20180425_133411

screenshot_20180425_133549

PR #679, opened 2018-06-18
removed psutil as a requirement

codingCoffee:
Works on linux. I need someone to test the following piece of code on MacOS and check if it correctly returns the total physical memory (RAM) in MBs in python.

```python
import os
(os.sysconf('SC_PAGE_SIZE') * os.sysconf('SC_PHYS_PAGES')) / (1024 * 1024)
``` 

Platform to test before merge
- [x] MacOS High Sierra >= 10.13 with python >= 3.6.4

### DIFF
--- a/bench/config/nginx.py
+++ b/bench/config/nginx.py
@@ -215,7 +215,6 @@ def get_error_pages():
 
 def get_limit_conn_shared_memory():
 	"""Allocate 2 percent of total virtual memory as shared memory for nginx limit_conn_zone"""
-	import psutil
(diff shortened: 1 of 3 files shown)


codingCoffee:
confirmed with @surajshetty3416 on MacOS High Sierra 10.13.5 and python 3.6.4

achillesrasquinha:
Does this also mean the C build would be required?

Also! Very impressive. Never knew about `os.sysconf`.

codingCoffee:
C build ??

codingCoffee:
Well yea, because of pillow and other dependencies, which too I'll try and remove

PR #685, opened 2018-06-26
Revert "get frappe_theme for foundation"

pratu16x7:
Reverts frappe/bench#684

`frappe_theme` removed as a requirement for `foundation`.


codingCoffee:
Merge-able when `frappe_theme` is removed from here https://github.com/erpnext/foundation/blob/master/requirements.txt

pratu16x7:
@codingCoffee done

PR #705, opened 2018-07-12
Autocompletion for bench (zsh and bash)

adityahase:
- Supports bench and frappe commands completion
- Supports --site completion (only lists sites)


adityahase:
Visit these [discuss](https://discuss.erpnext.com/t/command-line-completion-for-bench-commands/38405) [threads](https://discuss.erpnext.com/t/autocomplete-for-bench-and-how-to-use-it/38833) for help.

PR #709, opened 2018-08-21
Allow user to set single wildcard certificate for all domains

vjFaLk:
This is for a scenario when you have a single wildcard certificate for multiple domains. By default the nginx generation logic will only set the certs for domains that match the one set as the wildcard domain. 

However, by doing this, when you just set the domain as "*", the domain variable in the next line is set as empty string, which then is used in an `endswith`, which will always resolve `True`.


vjFaLk:
@codingCoffee Anything on this?

codingCoffee:
file permissions @vjFaLk !

vjFaLk:
@codingCoffee LOL, will fix and update

vjFaLk:
@codingCoffee Donezo!

PR #720, opened 2018-09-17
Force install pdfkit for new environments

### LINKED ISSUE #719: New Environment requires PDFKit by Force

achillesrasquinha:
Fixes #719


codingCoffee:
Already exists in exists in https://github.com/frappe/frappe/blob/develop/setup.py#L54

achillesrasquinha:
For some reason, it breaks during env migration. Check the list of installations prior to setup env

codingCoffee:
So I faced this error today while migrating. This should technically fix the issue, but it doesn't ! :/

sagarvora:
Not needed anymore, fixed in #744

PR #723, opened 2018-09-20
[fix] reload(): use importlib for python3

Thunderbottom:
`reload()` is an inbuilt function in python2 but has been moved to `importlib` since python3.4.

source: https://docs.python.org/3/library/importlib.html#importlib.reload

`Signed-off-by: Chinmay Pai <chinmaydpai@gmail.com>`


Thunderbottom:
I did.

Issue #730, opened 2018-10-22
GitPython Deadlock


dottenbr:
We have a GitPython version deadlock.

bench 4.1.0 has requirement GitPython=0.3.2.1

<img width="716" alt="git python 2" src="https://user-images.githubusercontent.com/9136094/47286847-ffd70280-d60d-11e8-8e73-d704d8674bbd.png">

frappe 10.1.54 has requirement GitPython 2.1.11

<img width="714" alt="git python 1" src="https://user-images.githubusercontent.com/9136094/47286874-11200f00-d60e-11e8-8084-eb75da5a8f64.png">

dottenbr:
The culprit:

![image](https://user-images.githubusercontent.com/9136094/47302053-21e87900-d63e-11e8-9f9f-a92d60aca956.png)

PR #731, opened 2018-10-22
Remove version requirements for GitPython

dottenbr:
https://github.com/frappe/bench/issues/730

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,4 +5,4 @@ requests
 honcho
 python-crontab
 semantic_version
-GitPython==0.3.2.1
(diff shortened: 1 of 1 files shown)


dottenbr:
@codingCoffee or @saurabh6790 can either of you look into this? As you see in https://github.com/frappe/bench/issues/730 this is causing problems that I believe we should work out before tomorrow's developer sprint has more people facing this?

codingCoffee:
Closing this, have pinned down the GitPython version. in https://github.com/frappe/bench/pull/733

PR #739, opened 2018-10-24
Provision to add custom rq queues

sharathc3:
Pull-Request

- [x] Have you followed the guidelines in our Contributing document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](../pulls) for the same update/change?
- [x] Have you lint your code locally prior to submission?
- [x] Have you successfully run tests with your changes locally?
- [x] Does your commit message have an explanation for your changes and why you'd like us to include them?
- [ ] Docs have been added / updated
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Did you modify the existing test cases? If yes, why?

---

What type of a PR is this? 

- [ ] Changes to Existing Features
- [ ] New Feature Submissions
- [ ] Bug Fix
- [x] Breaking Change
Default Queue definition in background jobs is moved to commom site config.
Without the migration, the existing setups will break.

--- 

- Motivation and Context (What existing problem does the pull request solve):
Queues in RQ was a hard coded into the system.
Now the hard coded values are removed and provision to add them via common_site config is added.

PR #740, opened 2018-10-25
Read Queue details from config into supervisor template

sharathc3:
Build queue worker details from common site config


barredterra:
Closing this. See https://github.com/frappe/bench/pull/739#issuecomment-787897515 for an explanation.

PR #744, opened 2018-11-07
fix: install app instead of requirements.txt

codingCoffee:
pip installs the frappe apps instead of just their requirements.txt
It needs to be done so as to prevent update the app version for pip and
to prevent conflits for the dependencies in requirements.txt

### DIFF
--- a/bench/utils.py
+++ b/bench/utils.py
@@ -176,7 +176,6 @@ def setup_env(bench_path='.', python = 'python'):
 	exec_cmd('virtualenv -q {} -p {}'.format('env', python), cwd=bench_path)
 	exec_cmd('{} -q install --upgrade pip'.format(pip), cwd=bench_path)
 	exec_cmd('{} -q install wheel'.format(pip), cwd=bench_path)
-	# exec_cmd('{pip} -q install https://github.com/frappe/MySQLdb1/archive/MySQLdb-1.2.5-patched.tar.gz'.format(pip), cwd=bench_path)
 	exec_cmd('{} -q install six'.format(pip), cwd=bench_path)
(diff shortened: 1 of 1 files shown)


codingCoffee:
This should prevent warning messages like so:
```
frappe 10.1.46 has requirement dropbox==7.3.1, but you'll have dropbox 9.1.0 which is incompatible.
```

We never install the frappe apps, since they are all installed in editable mode. Hence there is no need to install them again and again. However, since they are never reinstalled, `pip` only remembers the old locked down dependency of the `frappe apps`, hence we get error messages as shown above. So this is 1 way to fix the problem.

codingCoffee:
Also should take care of #720

bench/utils.py
```diff
@@ -430,9 +429,9 @@ def update_requirements(bench_path='.'):
 	bench_req_file = os.path.join(os.path.dirname(bench.__path__[0]), 'requirements.txt')
 	install_requirements(pip, bench_req_file)
 
+	from .app import install_app
```
sagarvora: maybe you should use absolute imports to prevent ambiguity 
(refer https://chrisyeh96.github.io/2017/08/08/definitive-guide-python-imports.html#absolute-vs-relative-import)
```suggestion
	from bench.app import install_app
```

rmehta:
If this is tested, merge it

sagarvora:
Hey @codingCoffee,

I tested `install_app` instead of `install_requirements` manually. I pinned `urllib` to an incompatible version to test whether error shows up in the console.

And it did:
```
requests 2.19.1 has requirement urllib3<1.24,>=1.21.1, but you'll have urllib3 1.20 which is incompatible.
```

This still doesn't solve the issue of having different dependencies for different python versions, but if any inconsistencies happen, they won't be silenced.

PR #753, opened 2018-12-12
 feat(CircleCI): tests shall not fail! ✔️

sagarvora:


README.md
```diff
@@ -4,7 +4,9 @@
     <h2>Frappe Bench</h2>
 </div>
 
-[![Build Status](https://travis-ci.org/frappe/bench.svg?branch=master)](https://travis-ci.org/frappe/bench)
+**Circle CI:** [![Circle CI](https://circleci.com/gh/frappe/bench.svg?style=svg)](https://circleci.com/gh/frappe/bench)
+
+**Travis CI:** [![Travis CI](https://travis-ci.org/frappe/bench.svg?branch=master)](https://travis-ci.org/frappe/bench)
```
codingCoffee: Remove this line. Not needed, as we shall discontinue with Travis for now.

bench/tests/test_init.py
```diff
@@ -78,8 +78,8 @@ def test_new_site(self):
 	def new_site(self, site_name):
 		new_site_cmd = ["bench", "new-site", site_name, "--admin-password", "admin"]
 
-		# set in travis
-		if os.environ.get("TRAVIS"):
+		# set in CI
+		if os.environ.get('CI'):
```
codingCoffee: `if os.environ.get("TRAVIS") or os.environ.get('CI'):` would be better I guess, in case we do go back.

bench/tests/test_init.py
```diff
@@ -186,7 +186,7 @@ def drop_site(self, site_name, archived_sites_path=None):
 		if archived_sites_path:
 			drop_site_cmd.extend(['--archived-sites-path', archived_sites_path])
 
-		if os.environ.get('TRAVIS'):
+		if os.environ.get('CI'):
```
codingCoffee: `if os.environ.get("TRAVIS") or os.environ.get('CI'):` would be better I guess, in case we do go back.

README.md
sagarvora: fixed.

PR #758, opened 2019-01-15
fix(release): don't fast-forward while merging

sagarvora:
as seen in release bench.


netchampfaris:
@sagarvora What does this mean? What is the behaviour before this patch? What is it after?

PR #786, opened 2019-04-09
feat: Remove hard-coded URLs, improve 'install_requires'

PMojito:
**Summary**

1. Improves setup.py by distinguishing between install_requires, and requirements files.  Add 3 different install_requires files, depending on the version of Python.
2. Removed hard-coded references to GitHub.  Replaced with a configuration file.

**Setup.py and Requirements**
There is a difference between install_requires, and Requirements Files.
 https://packaging.python.org/discussions/install-requires-vs-requirements/

- install_requires are _minimal_ requirements for a package to function.  They do not specify version numbers, unless a particular version is known to not work.
- Requirements Files are used for production installation and Long Term Support.  So that you always install exactly the same dependency versions.

I removed the name "requirements.txt" from setup.py, because that name is misleading.  I replaced with 3 new files:

- install_requires_2.txt.  This is for Python 2 installation.
- install_requires_3.txt.  This is for Python 3 installation.
- install_requires_3_5.txt.  This is for Python 3.5 and later (_Python 3.5 replaces virtual_env with venv.  So it has different dependencies_.)

I've also included a file named **requirements_LTS_v10.txt**.  The code does not reference this.  You do that during a pip install.  I've added comments inside the file, showing how to do that.

**Configuration-Driven URLs**
Throughout the Bench code, there were hard-coded references:
- https://github.com/frappe
- https://github.com/frappe/frappe.git
- ...and more.

What if we want to move our repository off GitHub, and to another host?  (i.e. GitLab)
What if individual developers want the Bench code pointing at their forks of Frappe and ERPNext?

My code tries to solve this problem.  I've added a TOML configuration file:
  **./bench/config/gitconfig.toml**

All the URLs and paths are now stored there!  So there's just 1 place to look for paths to GitHub.  You can edit the configuration file to meet your own needs.  If bench moves off GitHub in the future, it's easy to adjust.

Please let me know if you have any questions!
Brian


PMojito:
Closing due to lack of interest/feedback.  Will continue to maintain my own forked version.

PR #800, opened 2019-05-08
Fix: deprecated squash_actions

barredterra:
Fix deprecated squash actions according to : https://docs.ansible.com/ansible/devel/porting_guides/porting_guide_2.7.html#using-a-loop-on-a-package-module-via-squash-actions

### Changes

```yaml
- name: Install packages
  yum:
    state: present
    name: "{{ item }}"
  with_items: 
    - package_1
    - package_2
```

becomes

```yaml
- name: Install packages
  yum:
    state: present
    name:
      - package_1
      - package_2
```

### Effect

This will remove warnings like this one:

```
TASK [wkhtmltopdf : install base fonts] ****************************************
[DEPRECATION WARNING]: Invoking "yum" only once while using a loop via 
squash_actions is deprecated. Instead of using a loop to supply multiple items 
and specifying `name: "{{ item }}"`, please use `name: ['libXrender', 
'libXext', 'xorg-x11-fonts-75dpi', 'xorg-x11-fonts-Type1']` and remove the 
loop. This feature will be removed in version 2.11. Deprecation warnings can be
 disabled by setting deprecation_warnings=False in ansible.cfg.
```

### DIFF
--- a/playbooks/macosx.yml
+++ b/playbooks/macosx.yml
@@ -8,12 +8,13 @@
   tasks:
     # install pre-requisites
     - name: install prequisites
-      homebrew: name={{ item }} state=present
-      with_items:
-        - cmake
-        - redis
-        - mariadb
-        - nodejs
+      homebrew:
+        state: present
+        name:
+          - cmake
+          - redis
+          - mariadb
+          - nodejs
 
     # install wkhtmltopdf
     - name: cask installs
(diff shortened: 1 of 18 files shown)


gavindsouza:
There were conflicts at the time I checked out this PR a while back. Over time we merged https://github.com/frappe/bench/pull/880 that fixes the same. @barredterra thank you for your contribution but I'm closing this PR now.

PR #820, opened 2019-07-22
fix: validate branch before update

saurabh6790:


rdraju:
This will cause issues with bench update on projects where erpnext is not used. Bench & Frappe should be independent of applications built on them.

PR #844, opened 2019-10-05
chore: merge develop into master

gavindsouza:


bench/utils.py
```diff
@@ -25,21 +27,27 @@ def get_env_cmd(cmd, bench_path='.'):
...
+		os.makedirs(path)
 
-	os.makedirs(path)
 	for dirname in folders_in_bench:
-		os.mkdir(os.path.join(path, dirname))
+		try:
+			os.makedirs(os.path.join(path, dirname))
+		except OSError as e:
+			if e.errno != os.errno.EEXIST:
+				pass
```
gavindsouza: else Raise

bench/commands/setup.py
```diff
@@ -124,6 +124,17 @@ def setup_requirements():
 	update_requirements()
 	update_npm_packages()
 
+@click.command('manager')
```
gavindsouza: `master` branch has `def setup_manager` that does the same

PR #863, opened 2019-11-26
fix: don't update bench dependencies in env

gavindsouza:
added: user flag in `install_requirements` to install requirements at user level (specific to bench requirements)

### DIFF
--- a/bench/utils.py
+++ b/bench/utils.py
@@ -427,7 +427,12 @@ def set_default_site(site, bench_path='.'):
 
 def update_requirements(bench_path='.'):
 	print('Updating Python libraries...')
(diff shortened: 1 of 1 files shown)


BuddhiAbeyratne:
what is the proposed solution for people who already have bench installed in a venv ? as bench update is broken fo us right now as we are getting 
`ERROR: Can not perform a '--user' install. User site-packages are not visible in this virtualenv.`

for the moment ive manually set user=false inside of utils.py to continue working

gavindsouza:
@BuddhiAbeyratne Thanks for sharing this issue. I've opened https://github.com/frappe/bench/pull/866 that should fix this issue. Testing it right now

BuddhiAbeyratne:
awesome thanks for the quick fix

PR #876, opened 2019-12-15
fix(get-app): handle existing directory and other formatting changes

### LINKED ISSUE #496: [Feature Requests] Add cache and dependencies checker to Bench
While installing bench via the steps mentioned in the readme file, I encountered two issue.

* Bench didn't check the dependencies (e.g. npm) required for it's installation. It ran the process of build and complained about missing dependency at a much later stage. Ideally bench should be intelligent enough to check the dependencies available in the system and advice (or if possible install the dependencies ) the user accordingly.

* While installing bench or frappe via bench, it would be great to have a cache for downloaded date which gets pulled either via git or some other sources (in future), so that at a time when the install process or build process fails, it would continue from the stuf where it stopped, without the need to re-download or pull stuff. There is also a problem where bench complains and fails if there is already a folder that was left with a broken download or pull earlier. It should advice or automatically clean any corrupt or irrelevant folders

I hope I have ellaborated the issue correctly.

### LINKED ISSUE #554: Bench get-app should handle pre-existing app directories
When you do get-app, if the folder already contains a directory by same name git will fail.
Bench should check whether thereis already a app directory and offer an option to user
- Delete the existing directory and and re-do a get-app
- Do a get-app except skip the cloning part

Use case - 
When a user does get-app , there may be a situation when get-app fails in middle of the operation.
when user tries to do a get-app again it complains about exisiting directory, this should be handled interactively by bench, without the need for deleting the directory manually and redoing get-app

Thunderbottom:
* if a directory for the application already exists, provide user with two options:

  * ask to remove the directory and reclone (overwrite)
  * ask if the user wants to reinstall the application

* separate `get_app_name` from `get_app`

* move building assets and postprocessing to `install_app` method

* make formatting changes to the code for flake8

* use `reload_module` from six instead of builtins

closes #496 
closes #554


bench/app.py
```diff
@@ -93,58 +95,64 @@ def remove_from_excluded_apps_txt(app, bench_path='.'):
...
+		# application directory already exists
+		# prompt user to overwrite it
+		if overwrite or click.confirm('''A directory for the application "{0}" already exists.
+Do you want to continue and overwrite it?'''):
+			shutil.rmtree(os.path.join(bench_path, 'apps', repo_name))
+		elif click.confirm('''Do you want to reinstall the existing application?''', abort=True):
+			app_name = get_app_name(bench_path, repo_name)
+			print("Reinstalling {0}".format(app_name))
+			install_app(app=app_name, bench_path=bench_path, verbose=verbose, build_asset_files=build_asset_files)
+			sys.exit(1)
```
gavindsouza: if the application is installed successfully, then why exit with status 1?

bench/commands/make.py
```diff
@@ -35,10 +35,11 @@ def init(path, apps_path, frappe_path, frappe_branch, no_procfile, no_backups,
 @click.argument('name', nargs=-1) # Dummy argument for backward compatibility
 @click.argument('git-url')
 @click.option('--branch', default=None, help="branch to checkout")
+@click.option('--overwrite', is_flag=True)
 def get_app(git_url, branch, name=None):
```
gavindsouza: missing `overwrite` keyword argument

bench/app.py
```diff
@@ -93,58 +95,64 @@ def remove_from_excluded_apps_txt(app, bench_path='.'):
...
 	repo_name = git_url.rsplit('/', 1)[1].rsplit('.', 1)[0]
-	logger.info('getting app {}'.format(repo_name))
 	shallow_clone = '--depth 1' if check_git_for_shallow_clone() else ''
 	branch = '--branch {branch}'.format(branch=branch) if branch else ''
 
+	if os.path.isdir(os.path.join(bench_path, 'apps', repo_name)):
+		# application directory already exists
+		# prompt user to overwrite it
+		if overwrite or click.confirm('''A directory for the application "{0}" already exists.
+Do you want to continue and overwrite it?'''):
```
gavindsouza: missing format string for app name

bench/app.py
```diff
@@ -93,58 +95,64 @@ def remove_from_excluded_apps_txt(app, bench_path='.'):
...
+
+	logger.info('Getting app {0}'.format(repo_name))
+	exec_cmd("git clone {git_url} {branch} {shallow_clone} --origin upstream".format(
+		git_url=git_url,
+		shallow_clone=shallow_clone,
+		branch=branch),
+		cwd=os.path.join(bench_path, 'apps'))
+
+	app_name = get_app_name(bench_path, repo_name)
+	print("Installing {0}".format(app_name))
```
gavindsouza: Screenshot 2020-02-25 at 1 06 17 PM

Is that print statement really necessary? `(along with bench logger's outputs)`

PR #877, opened 2019-12-16
fix(bench): use gitpython instead of subprocess and minor fixes

Thunderbottom:
* use gitpython for git commands instead of using subprocess.
* rewrite pip commands using [`subprocess.check_call()`](https://pip.pypa.io/en/latest/user_guide/#using-pip-from-your-program).
* code formatting changes.


gavindsouza:
Please open this PR when ready.

PR #882, opened 2019-12-19
feat: skip build assets while doing get-app

shridarpatil:
Skip build assets while running bench get-app


bench/commands/make.py
```diff
@@ -35,10 +35,11 @@ def init(path, apps_path, frappe_path, frappe_branch, no_procfile, no_backups,
 @click.argument('name', nargs=-1) # Dummy argument for backward compatibility
 @click.argument('git-url')
 @click.option('--branch', default=None, help="branch to checkout")
-def get_app(git_url, branch, name=None):
+@click.option('--skip-assets', is_flag=True, default=True, help="Do not build assets")
```
gavindsouza: default behaviour should be to not skip assets. So changing default to False would be apt

PR #885, opened 2019-12-27
fix(easy_install): updated sequence of package updates

gavindsouza:
**fix:** updated sequence of package updates (prerequisites install issue in debian)
**fix:** os compatible paths
**chore:** breakdown install_bench in smaller functions


gavindsouza:
tested on ub18 and deb9.7

gavindsouza:
@jalajc I'll update the docs. thanks for pointing this out!

PR #898, opened 2020-01-13
Merge master into develop

barredterra:
Cannot work on `develop` as long as the differences are that big.


gavindsouza:
@barredterra does `develop` have any features that `master` doesn't currently?

Apart from there being conflicts and about 200 files to review, we're planning on discontinuing use of `develop` after #844

Thunderbottom:
dropped develop, the changes in there aren't required.

PR #903, opened 2020-01-17
fix: validate branch in bench update

ckosiegbu:
bench update assumes that erpnext is an installed app in the bench. This is a wrong assumption. This work around checks for the intersection of installed apps with  the list ['frappe', 'erpnext'] and only attempts to get the current branch of apps that occur in the intersection. This way, if erpnext is not installed (very likely for those using frappe only to build their own apps) there wont be an error when get_current_branch is called for erpnext. The need to perform the kind of check being performed in validate_branch is a strong case for considering apps being able to add "bench update" hooks in their hooks.py

See this topic https://discuss.erpnext.com/t/bench-update-error-no-such-file-or-directory-apps-erpnext/56932?u=chude_osiegbu

Pull-Request

- [ ] Have you followed the guidelines in our Contributing document?
- [ ] Have you lint your code locally prior to submission?

---

What type of a PR is this? 

- [ ] New Feature Submissions
- [x] Bug Fix
- [ ] Breaking Change

--- 

- Motivation and Context (What existing problem does the pull request solve):
- Related Issue: 
- Screenshots (if applicable, remember, a picture tells a thousand words): 

**Please don't be intimidated by the long list of options you've fill. Try to fill out as much as you can. Remember, the more the information the easier it is for us to test and get your pull request merged** :grin:

### DIFF
--- a/bench/app.py
+++ b/bench/app.py
@@ -423,7 +423,10 @@ def get_apps_json(path):
 		return json.load(f)
 
 def validate_branch():
-	for app in ['frappe', 'erpnext']:
+	installed_apps = get_apps()
+	check_apps = ['frappe', 'erpnext'] #this is a dirty work-around. bench update should not assume that erpnext is installed
+	intersection_apps = list(set(installed_apps) & set(check_apps))
+	for app in intersection_apps:
 		branch = get_current_branch(app)
 
 		if branch == "master":
--- a/bench/app.py
+++ b/bench/app.py
@@ -423,14 +423,22 @@ def get_apps_json(path):
 		return json.load(f)
 
 def validate_branch():
-	installed_apps = get_apps()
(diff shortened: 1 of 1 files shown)


gavindsouza:
@ckosiegbu the PR and commit names don't conform with our standards. I'll rename them and merge it soon

ckosiegbu:
Hi @gavindsouza the bug this correction was supposed to correct still exists because this line:

> intersection_apps = installed_apps and check_apps

does not return a proper intersections. Instead it returns the last set in the statement. See example below:

> >>> a = {'frappe'}
> >>> b = {'erpnext', 'frappe'}
> >>> c = a and b
> >>> print (c)
> {'erpnext', 'frappe'}

I think we should revert to:

> intersection_apps = list(set(installed_apps) & set(check_apps))

or 

> intersection_apps = set(installed_apps) & set(check_apps)

Issue #777, opened 2019-02-19
Install Frappe from local filesystem


Dremor:
Issue: Feature Request

**Is your feature request related to a problem? Please describe.**
I'm currently trying to make an alpine-based Docker ERPNext image (I know there is already some other Docker image, but I do this for fun). As such, I could have to do some change to Frappe or ERPNext in order to make it work on this "exotic" base.

I started to work using git submodules (and patches), in order to keep as close as possible to upstream (as seen here : https://lab.dremor.info/egeorget/erpnext-docker/tree/feature/research). But Bench do not allow to install Frappe from a local filesystem (or a local git repository in this case), which prevent this project topology to work.

**Describe the solution you'd like**
I would like to be able to get Frappe (or any bench app) from a local directory (being a local filesystem or a local git repository).

gavindsouza:
Possible solution:

move the app to `{bench-folder}/apps` and add it to `apps.txt` under `{bench-folder}/sites`

then proceed to `bench --site {site-name} {app}`

gavindsouza:
Resolved in #900

zeelrupapara:
This solution works for me:
I used the `file` protocol to get the app in the bench (for local file transfer)
`bench get-app file:///home/custom-app`

varun-krishnamurthy:
This is not working anymore. 

frappe@3da05a912878:/workspace/development/frappe-bench/apps$ bench --version
5.17.2

frappe@3da05a912878:/workspace/development/frappe-bench/apps$ bench get-app file://tmp/hrms/
ERROR: file: not found in frappe or erpnext
Traceback (most recent call last):
  File "/home/frappe/.bench/bench/utils/__init__.py", line 437, in fetch_details_from_tag
    org, repo = org_repo
ValueError: too many values to unpack (expected 2)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/frappe/.local/bin/bench", line 8, in <module>
    sys.exit(cli())
  File "/home/frappe/.bench/bench/cli.py", line 132, in cli

PR #900, opened 2020-01-15
chore: add travis, deepsource

gavindsouza:
Closes #944

PR #913, opened 2020-02-05
feat: get-app from folder in filesystem

gavindsouza:
Frappe apps can be installed from existing repositories in the filesystem

Screenshot 2020-02-05 at 1 05 19 PM

Closes https://github.com/frappe/bench/issues/777

### DIFF
--- a/bench/app.py
+++ b/bench/app.py
@@ -1,31 +1,28 @@
-from __future__ import print_function
-import os
-from .utils import (exec_cmd, get_frappe, check_git_for_shallow_clone, build_assets,
-	restart_supervisor_processes, get_cmd_output, run_frappe_cmd, CommandFailedError,
-	restart_systemd_processes, log)
-from .config.common_site_config import get_config
+import json, logging, os, re, shutil, subprocess, sys
 
-import logging
 import requests
 import semantic_version
-import json
-import re
-import subprocess
+
 import bench
-import sys
-import shutil
+from bench.config.common_site_config import get_config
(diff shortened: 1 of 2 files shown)


bench/app.py
```diff
@@ -1,55 +1,56 @@
-from __future__ import print_function
-import os
-from .utils import (exec_cmd, get_frappe, check_git_for_shallow_clone, build_assets,
-	restart_supervisor_processes, get_cmd_output, run_frappe_cmd, CommandFailedError,
-	restart_systemd_processes)
-from .config.common_site_config import get_config
-
-import logging
+import json, logging, os, re, shutil, subprocess, sys
```
Thunderbottom: i dont think we should be going anti-pattern and doing multiple imports on single line.

bench/app.py
```diff
@@ -82,47 +85,61 @@ def add_to_excluded_apps_txt(app, bench_path='.'):
...
+				cwd=os.path.join(bench_path, 'apps'))
+
+	return repo_name
+
+
+def get_app(url, branch=None, bench_path='.', skip_assets=False, verbose=False, postprocess=True):
+	"""Installs Frappe app from url (git repo or file system) in bench env"""
+
+	try:
+		if os.path.exists(url):
```
Thunderbottom: instead of doing all this, can we add a flag `--local` to handle local files?

gavindsouza: any specific reason for adding a `local` flag?

Thunderbottom: its easier for the user to just add another flag than to add more complications by checking if the path exists locally before trying to clone from remote.

gavindsouza:
Resolved in #900

PR #918, opened 2020-02-07
fix: setup cron job for renewing ssl certificate

gavindsouza:
**What this does:** sets cron job under `root` user set for 1st of every month

**Reason:** `setup_crontab` failed silently on multiple tries on new Ubuntu 18.04 droplet and macOS 18.7.0

**Package Ref:** [python-crontab](https://gitlab.com/doctormo/python-crontab)


gavindsouza:
tested on mac and Ubuntu, py2.7 and py3.7

Thunderbottom:
looks good to me.

can you change the requirements to reflect `python-crontab` and not `python_crontab`. because even though the one we have right now works, i think it's better to use the canonical name from pypi instead of relying on [safe-magic](https://mail.python.org/pipermail/distutils-sig/2011-August/017936.html)

PR #919, opened 2020-02-10
fix: dont abort script if not overwriting scripts

gavindsouza:
currently, running of `bench ` or `bench setup production` creates nginx conf files and asks for confirmation before overwriting them. in case user doesn't want to overwrite files, `setup production` and `lets-encrypt` is exitted abruptly and nginx isn't started again after. 

this happens because click abort was being used which sends a script termination signal in case of `n` instead of returning control to parent function.

### DIFF
--- a/bench/config/nginx.py
+++ b/bench/config/nginx.py
@@ -1,14 +1,23 @@
-import os, json, click, random, string, hashlib
-from bench.utils import get_sites, get_bench_name, exec_cmd
+# imports - standard imports
+import hashlib
+import json
+import os
+import random
+import string
+
+# imports - third party imports
+import click
 from six import string_types
-from six.moves import input
+
+# imports - module imports
+from bench.utils import exec_cmd, get_bench_name, get_sites
 
 
 def make_nginx_conf(bench_path, yes=False):
 	conf_path = os.path.join(bench_path, "config", "nginx.conf")
 
 	if not yes and os.path.exists(conf_path):
-		if input('nginx.conf already exists and this will overwrite it. Do you want to continue? [y/N]').lower() == 'n':
+		if not click.confirm('nginx.conf already exists and this will overwrite it. Do you want to continue?'):
 			return
 
 	from bench import env
--- a/bench/config/nginx.py
+++ b/bench/config/nginx.py
@@ -1,6 +1,5 @@
 # imports - standard imports
 import hashlib
-import json
 import os
 import random
 import string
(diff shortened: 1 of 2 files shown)


bench/config/nginx.py
```diff
@@ -1,49 +1,49 @@
...
 
+def make_nginx_conf(bench_path, yes=False):
 	conf_path = os.path.join(bench_path, "config", "nginx.conf")
 	if not yes and os.path.exists(conf_path):
-		click.confirm('nginx.conf already exists and this will overwrite it. Do you want to continue?',
-			abort=True)
-
-	with open(conf_path, "w") as f:
-		f.write(nginx_conf)
+		if input('nginx.conf already exists and this will overwrite it. Do you want to continue? [y/N]').lower() == 'y':
```
saurabh6790: Where is the code that restarts nginx? If that code is in another method then you can just check

``` 
if input('nginx.conf already exists and this will overwrite it. Do you want to continue? [y/N]').lower() == 'N':
    return
```

Instead of moving code under if block

gavindsouza: `setup_letsencrypt` calls `run_certbot_and_setup_ssl` which calls `make_nginx_conf` at [L79](https://github.com/frappe/bench/blob/master/bench/config/lets_encrypt.py#L79) and restarts nginx at [L80](https://github.com/frappe/bench/blob/master/bench/config/lets_encrypt.py#L80)

yes that's more apt. thanks 👍

bench/config/nginx.py
```diff
@@ -1,8 +1,16 @@
...
 from bench.utils import get_sites, get_bench_name, exec_cmd
 from six import string_types
+from six.moves import input
+
 
 def make_nginx_conf(bench_path, yes=False):
+	conf_path = os.path.join(bench_path, "config", "nginx.conf")
+
+	if not yes and os.path.exists(conf_path):
+		if input('nginx.conf already exists and this will overwrite it. Do you want to continue? [y/N]').lower() == 'n':
```
Thunderbottom: instead of this we can still use click without `abort=True` and then indent all the code inside this.

anything that is not 'n' will still be rendered as a yes with this.

gavindsouza: Something like https://github.com/frappe/bench/commit/ba90c9fd7c919add30ee94f9e3f85540d26948c0 but with `click.confirm`?

Thunderbottom: https://click.palletsprojects.com/en/7.x/prompts/#confirmation-prompts

so something like: 
```python3
if click.confirm("Press enter to continue: "):
	setup_nginx()
	...
```

PR #929, opened 2020-02-20
chore: fix easy_install

Thunderbottom:
import print from `__future__` for python2 to use print's `file=` param


gavindsouza:
first commit in https://github.com/frappe/bench/pull/928

Thunderbottom:
no, this doesnt work, why was it merged?

PR #934, opened 2020-02-25
fix: show warning if bench is not installed via PYPI

gavindsouza:
Screenshot 2020-03-16 at 5 44 26 PM

### DIFF
--- a/bench/utils.py
+++ b/bench/utils.py
@@ -8,6 +8,7 @@ from six import iteritems
 from six.moves.urllib.parse import urlparse
 
 import bench
+from bench import env
 
 
 class PatchError(Exception):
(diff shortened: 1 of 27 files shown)


bench/utils.py
```diff
@@ -1171,6 +1171,15 @@ def migrate_env(python, backup=False):
 		raise
 
 
+def is_dist_editable(dist):
+	"""Is distribution an editable install?"""
+	for path_item in sys.path:
+		egg_link = os.path.join(path_item, dist + '.egg-link')
```
gavindsouza: .egg-link file exists in site-packages if installed in editable mode. eg: `/Library/Python/2.7/site-packages/bench.egg-link`

PR #950, opened 2020-03-05
feat(procfile): dont add redis to procfile if skipping config gen

Thunderbottom:


revant:
Tried locally.

create new bench without redis. works as expected

PR #952, opened 2020-03-11
chore: remove updating "bench tool" from update

gavindsouza:
- removes `bench_update_on_update`
- removes `restart_update`
- removes `validate_pillow_dependencies`
- adds setup_env: find venv in current python and failure log 
- style fixes 

_Pending:_ docs update


Alchez:
@gavindsouza, are the bench updates going to be done through PyPI instead?

gavindsouza:
@Alchez yes!

PR #954, opened 2020-03-13
chore: update README for docker

revant:
Changed docker section to reference development and production docker images


revant:
this will close https://github.com/frappe/frappe_docker/issues/102

PR #956, opened 2020-03-19
(docs): Added restore command to bench_usage.md

pm-at:
Might be helpful for people trying to restore directly with mysql.


docs/bench_usage.md
```diff
@@ -79,6 +79,7 @@ These commands belong directly to the bench group so they can be invoked directl
  - **renew-lets-encrypt**: Renew Let's Encrypt certificate for site SSL.
  - **backup**: Backup single site data. Can be used to backup files as well.
  - **backup-all-sites**: Backup all sites in current bench.
+ - **restore**: Restore a site's database and optionally, its public and private file backups.
```
gavindsouza: The `restore` command is a part of [frappe.commands.site](https://github.com/frappe/frappe/blob/develop/frappe/commands/site.py#L98) along with a lot of other commands. 

Maybe we should add a section at the bottom for frappe's documentation too.

PR #982, opened 2020-05-04
fix: SocketIO HTTPs bug

SaiFi0102:
Socket IO messages are not received on HTTPs when the configuration:
`ssl_ecdh_curve` is set to `secp384r1`

Setting it to `auto` solves it according to Telegram Developer's Group message by Shadoyip

### DIFF
--- a/bench/config/templates/nginx.conf
+++ b/bench/config/templates/nginx.conf
@@ -41,7 +41,7 @@ server {
 	ssl_stapling_verify on;
 	ssl_protocols TLSv1.2 TLSv1.3;
 	ssl_ciphers EECDH+AESGCM:EDH+AESGCM;
-	ssl_ecdh_curve secp384r1;
+	ssl_ecdh_curve auto;
 	ssl_prefer_server_ciphers on;
 	{% endif %}
 


Thunderbottom:
mistakenly clicked on the close button before commenting. i think this really isnt an issue as of now. this defaults to `prime256v1`, which even though secure is not as fast as `secp384r1`.

there is some other misconfiguration that causes socketio to fail and not this, since most (all modern?) clients do support `secp384r1`, and is recommended to be used as the default cipher for both speed and security.

Thunderbottom:
just checked with @revant, socketio works without these changes too.

could you tell us what exactly did not work for you?

Thunderbottom:
could you try setting the socketio block to this (without making the changes from this PR):

```nginx
location /socket.io {
		proxy_http_version 1.1;
		proxy_set_header Upgrade $http_upgrade;
		proxy_set_header Connection "upgrade";
		proxy_set_header X-Frappe-Site-Name $http_host;
		proxy_set_header Origin $scheme://$http_host;
		proxy_set_header Host $host;

		proxy_pass http://{{ bench_name }}-socketio-server;
	}
```

and let me know if it works?

pm-at:
I'm facing this issue too. My results as below:

> basically setting the `X-Frappe-Site-Name` header to `$http_host`

Works after `nginx reload` but breaks after `supervisorctl reload`

Setting `ssl_ecdh_curve` to `auto` works after `nginx reload` as well as `supervisorctl reload`

This thing broke silently so didn't notice until a user complained.

revant:
> just checked with @revant, socketio works without these changes too.

I only checked chat. It was working.

Opening same doctype by two users doesn't show message anymore.

sagarvora:
Here's what worked for me. Node JS v8 is what gets installed with easy install script. I just upgraded to v12 (latest stable) and the SSL error went away.

sagarvora:
> Here's what worked for me. Node JS v8 is what gets installed with easy install script. I just upgraded to v12 (latest stable) and the SSL error went away.

@SaiFi0102  can you try this please?

SaiFi0102:
Apparently, I test both Node v8 and Node v12... However with Node v12 and HTTPS and Frappe v12 it's working just fine.

There was another problem that caused issue with reloading document when the document was modified on another session which I solved here https://github.com/frappe/frappe/pull/11137

adityahase:
@SaiFi0102 Can we close this now?

SaiFi0102:
Please go ahead

Issue #993, opened 2020-05-11
Get-app fails with SSH remote URLs


Alchez:
Issue: Bug report

**Do the checklist before filing an issue:**
 - [x] Can you replicate the issue?
 - [ ] Is this something you can debug and fix? Send a pull request! Bug fixes and documentation fixes are welcome

**Describe the bug** :chart_with_downwards_trend:
On running a `get-app` on a non-standard Frappe application, bench tries to find the application inside the Frappe organization.

**To Reproduce** :page_with_curl:
1. Inside a bench folder, run `bench get-app git@github.com:<org>/<app>.git`
2. See error

**OS (please complete the following information):** :cyclone:
- [x] Linux: `Ubuntu 20.04 LTS`
- [ ] macOS: `version`
- [ ] Windows `version`
- [ ] Others? `haros:distro:version`

**Version Information**

- Bench Branch: `develop` _(Only master is supported)_
- Frappe Version: 12.3.0
- ERPNext Version: 12.5.2

**Possible Solution** :bookmark_tabs:
In the `get_app` function, bench tries to check if the provided argument is a URL by checking for schemes like `http` or `https`. This fails for SSH URLs. A possible fix could be to include git links.

PR #995, opened 2020-05-12
fix(get-app): allow ssh urls

gavindsouza:
Closes https://github.com/frappe/bench/issues/993

### DIFF
--- a/bench/app.py
+++ b/bench/app.py
@@ -100,7 +100,9 @@ def remove_from_excluded_apps_txt(app, bench_path='.'):
 
 def get_app(git_url, branch=None, bench_path='.', skip_assets=False, verbose=False, postprocess=True, overwrite=False):
 	if not os.path.exists(git_url):
-		if not check_url(git_url, raise_err=False):
+		if git_url.startswith('git@'):
+			pass
+		elif not check_url(git_url, raise_err=False):
 			orgs = ['frappe', 'erpnext']
 			for org in orgs:
 				url = 'https://api.github.com/repos/{org}/{app}'.format(org=org, app=git_url)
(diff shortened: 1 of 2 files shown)


gavindsouza:
Tested manually:
-  without explicitly specifying URL (frappe, erpnext apps)
- using HTTP links (http/https)
- using SSH links (git)

via tests:
- using local file paths
- using HTTP links

bench/app.py
```diff
@@ -100,7 +100,9 @@ def remove_from_excluded_apps_txt(app, bench_path='.'):
 
 def get_app(git_url, branch=None, bench_path='.', skip_assets=False, verbose=False, postprocess=True, overwrite=False):
 	if not os.path.exists(git_url):
-		if not check_url(git_url, raise_err=False):
+		if git_url.startswith('git@'):
+			pass
+		elif not check_url(git_url, raise_err=False):
```
Thunderbottom, line 105:
```suggestion
		if not git_url.startswith('git@') and not check_url(git_url, raise_err=False):
```

Alchez: Would it be better to handle it in the `check_url` function, which could just be calling new `check_ssh_url` and `check_https_url` functions?

Thunderbottom: sure, that works too

bench/app.py
```diff
@@ -100,7 +105,7 @@ def remove_from_excluded_apps_txt(app, bench_path='.'):
 
 def get_app(git_url, branch=None, bench_path='.', skip_assets=False, verbose=False, postprocess=True, overwrite=False):
 	if not os.path.exists(git_url):
-		if not check_url(git_url, raise_err=False):
```
Alchez, line 103: Not sure if this is already on your to-do, but we should remove the `check_url` function as well

gavindsouza: Not at the moment; There are a few dead functions floating around bench. We'll need a cleanup, we could even remove all the frappe < v5 support :P

Alchez: Oh, for sure

gavindsouza:
@Mergifyio backport master

PR #1007, opened 2020-05-26
fix: set referrer-policy header in nginx config

Thunderbottom:
sets the "Referrer-Policy" header to "strict-origin-when-cross-origin", with "same-origin" as a fallback

setting the referrer policy prevents sharing site context to external links, preventing cross-site hijacking or tab nagging.

ref: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referrer-Policy


surajshetty3416:
@Thunderbottom will this update already created nginx configs?

Thunderbottom:
no, this does not. requires a `bench setup nginx` to run.

surajshetty3416:
@Thunderbottom Can we add an info/warning message for users to update their nginx config?

Thunderbottom:
yeah, will have to see how exactly to do it.

gavindsouza (approved):

Issue #1016, opened 2020-06-08
Replace platform.dist() with distro.linux_distribution in install.py | platform.dist() not supported by python3.8


nikunj1222:
Issue: Easy Install

**Distro Information (Required)**

Ubuntu 20.04 LTS
And for all new distros in future

**Possible Solution**

To replace **Replace platform.dist() with distro.linux_distribution in install.py**

Use this module https://github.com/nir0s/distro for getting distro/system verion information.

I will try to cook something myself. But I am not a professional programmer. So if anybody can work on this request.

nikunj1222:
#1019 

Check this pull request to manage the installation on ubuntu 20.04

gavindsouza:
Fixed with https://github.com/frappe/bench/pull/1048

Issue #997, opened 2020-05-14
Easy Install Support for Ubuntu 20.04 LTS


kevinpthorne:
Issue: Feature Request

I had some fun attempting to do a production installation of bench on a fresh Ubuntu 20.04 LTS with Python 3.8 this morning. Yes, Ubuntu 20.04 is still [very VERY new](https://lists.ubuntu.com/archives/ubuntu-announce/2020-April/000256.html).

#### To reproduce
As stated in Bench README:
```bash
apt install python3-minimal build-essential python3-setuptools
wget https://raw.githubusercontent.com/frappe/bench/develop/install.py
python3 install.py --production
```

### `platform.dist()` removed in Python 3.8
Python 3.8 removed finally removed the deprecated `platform.dist()` method. This has been [deprecated since Python 2.6](https://docs.python.org/2/library/platform.html#platform.dist). Although changing `install.py:100` to
`current_dist = ('Ubuntu', '20.04', 'focal')`
easily fixed that issue. I grabbed that running `platform.dist()` in a Python 2 console.

### `libssl1.0-dev` unavailable in Ubuntu 20.04 LTS

The easy install script kept going for some time until it reached this:
```
TASK [common : install pdf prerequisites for Ubuntu >= 18.04] ************************************************************************************************************************************************************************
task path: /tmp/.bench/playbooks/roles/common/tasks/ubuntu.yml:33
<127.0.0.1> ESTABLISH LOCAL CONNECTION FOR USER: root
<127.0.0.1> EXEC /bin/sh -c 'echo ~root && sleep 0'
<127.0.0.1> EXEC /bin/sh -c '( umask 77 && mkdir -p "` echo /root/.ansible/tmp/ansible-tmp-1589483796.3580744-140108001331921 `" && echo ansible-tmp-1589483796.3580744-140108001331921="` echo /root/.ansible/tmp/ansible-tmp-1589483796.3580744-140108001331921 `" ) && sleep 0'
Using module file /usr/local/lib/python3.8/dist-packages/ansible/modules/packaging/os/apt.py
<127.0.0.1> PUT /root/.ansible/tmp/ansible-local-349933sshhs80/tmpgx2rvu9o TO /root/.ansible/tmp/ansible-tmp-1589483796.3580744-140108001331921/AnsiballZ_apt.py
[...]

PR #1017, opened 2020-06-09
fix(install): Replace platform.dist with distro.linux_distribution for PY3.8+ compatibility

nikunj1222:
Replace platform.dist() with distro.linux_distribution in install.py | platform.dist() not supported by python3.8

Distro Information (Required)

Python >= 3.5
Ubuntu 20.04 LTS
And for all new distros in future

Possible Solution

To replace Replace platform.dist() with **distro.linux_distribution** in install.py

Use this module https://github.com/nir0s/distro for getting distro/system verion information.

Install distro module

`
Sudo pip install distro

`

### DIFF
--- a/install.py
+++ b/install.py
@@ -11,13 +11,18 @@ import platform
 import warnings
 import datetime
 import importlib
-try:
-    importlib.import_module(distro)
-except ImportError:
-    import pip
-    pip.main(['install', distro])
-finally:
-    globals()[package] = importlib.import_module(distro)
+
+def install_and_import(package):
+    import importlib
+    try:
+        importlib.import_module(package)
+    except ImportError:
+        import pip
+        pip.main(['install', package])
+    finally:
+        globals()[package] = importlib.import_module(package)
+
+install_and_import('distro')
 
 tmp_bench_repo = os.path.join('/', 'tmp', '.bench')
 tmp_log_folder = os.path.join('/', 'tmp', 'logs')
--- a/install.py
+++ b/install.py
@@ -10,12 +10,14 @@ import shutil
 import platform
 import warnings
 import datetime
+import importlib
 try:
-  import distro
+    importlib.import_module(distro)
 except ImportError:
-  print("Trying to Install required module: distro")
-  os.system('python -m pip install distro')
-import distro
+    import pip
+    pip.main(['install', distro])
+finally:
+    globals()[package] = importlib.import_module(distro)
 
 tmp_bench_repo = os.path.join('/', 'tmp', '.bench')
 tmp_log_folder = os.path.join('/', 'tmp', 'logs')
--- a/install.py
+++ b/install.py
@@ -136,7 +136,6 @@ def install_prerequisites():
 	# pre-requisites for bench repo cloning
 	run_os_command({
 		'apt-get': [
-			'sudo apt-get update',
 			'sudo apt-get install -y git build-essential python3-setuptools python3-dev libffi-dev'
 		],
 		'yum': [
@@ -183,6 +182,10 @@ def install_package(package, package_name=None):
 
 
 def install_pip():
+	run_os_command({
+		'apt-get': 'sudo apt-get update',
+		'yum': 'sudo yum check-update'
+	})
 	if PY3:
 		install_package('pip3', 'python3-pip')
 	else:
--- a/install.py
+++ b/install.py
@@ -109,13 +109,8 @@ def get_distribution_info():
 
     # return distribution name and major version
 
-    if platform.system() == 'Linux':
-        if sys.version_info < (3, 7):
-            current_dist = platform.dist()
-        else:
-            current_dist = (distro.id(), distro.major_version())
-	
-            return (current_dist[0].lower(), current_dist[1].rsplit(".")[0])
+	if platform.system() == 'Linux':
+		return (distro.id(), distro.major_version())
 	
     elif platform.system() == 'Darwin':
 
--- a/install.py
+++ b/install.py
@@ -17,8 +17,7 @@ def install_and_import(package):
     try:
         importlib.import_module(package)
     except ImportError:
-        import pip
-        pip.main(['install', package])
+		subprocess.check_call('{0} -m pip install {1}'.format(sys.executable, package))
     finally:
         globals()[package] = importlib.import_module(package)
 
--- a/install.py
+++ b/install.py
@@ -12,16 +12,6 @@ import warnings
 import datetime
 import importlib
 
-def install_python_package(package):
-    try:
-        importlib.import_module(package)
-    except ImportError:
-	print("Trying to Install required module: " + package)
-	subprocess.check_call([sys.executable, "-m", "pip", "install", package])
-    finally:
-        globals()[package] = importlib.import_module(package)
-
-
 
 tmp_bench_repo = os.path.join('/', 'tmp', '.bench')
 tmp_log_folder = os.path.join('/', 'tmp', 'logs')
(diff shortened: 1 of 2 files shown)


install.py
```diff
@@ -97,7 +97,11 @@ def check_distribution_compatibility():
 def get_distribution_info():
 	# return distribution name and major version
 	if platform.system() == "Linux":
-		current_dist = platform.dist()
+        if (platform.python_version_tuple()[0] + platform.python_version_tuple()[1]) == '2.7':
```
nikunj1222, line 100: If the it is python version 2.7 use  platform.dist() for python3 use dist.linux_distribution()

nikunj1222:
The do think now the failure is coming from other dependencies...?

nikunj1222:
#1019 https://github.com/frappe/bench/pull/1019

Should be merged to solved the errors on python 3.7,3.6,3.8

nikunj1222:
#1016 #997 
Above issues can be solved with this pull request.

install.py
```diff
@@ -10,7 +10,19 @@
...
+    import importlib
+    try:
+        importlib.import_module(package)
+    except ImportError:
+        import pip
+        pip.main(['install', package])
+    finally:
+        globals()[package] = importlib.import_module(package)
+
+install_and_import('distro')
```
Alchez, line 25: You can replace all of this with a simple `import distro`, and add that package in bench's `requirements.txt` file

install.py
```diff
@@ -95,14 +106,23 @@ def check_distribution_compatibility():
...
-	elif platform.system() == "Darwin":
-		current_dist = platform.mac_ver()
-		return "macos", current_dist[0].rsplit('.', 1)[0]
+
+    # return distribution name and major version
+
+    if platform.system() == 'Linux':
+        if int(platform.python_version_tuple()[0]
+               + platform.python_version_tuple()[1]) <= 3.7:
+            current_dist = platfor.dist()
```
Alchez, line 115: You have a typo here

install.py
```diff
@@ -95,14 +106,23 @@ def check_distribution_compatibility():
...
-
-	elif platform.system() == "Darwin":
-		current_dist = platform.mac_ver()
-		return "macos", current_dist[0].rsplit('.', 1)[0]
+
+    # return distribution name and major version
+
+    if platform.system() == 'Linux':
+        if int(platform.python_version_tuple()[0]
+               + platform.python_version_tuple()[1]) <= 3.7:
```
Alchez, line 114: You can use the `sys` module here

```suggestion
        if sys.version_info < (3, 7):
```

install.py
```diff
@@ -95,14 +106,23 @@ def check_distribution_compatibility():
...
+            current_dist = platfor.dist()
+        else:
+            current_dist = \
+                distro.linux_distribution(full_distribution_name=True)
+
+            return (current_dist[0].lower(), current_dist[1].rsplit('.'
+                    )[0])
+    elif platform.system() == 'Darwin':
+
```
Alchez, line 123: Remove the extra line

install.py
gavindsouza: @Alchez `install.py` should be able to set up a frappe env with a single command on a fresh server. It should manage all the dependencies it needs. Although they share the same repo, `install.py` isn't a part of bench, so we don't need to update `requirements.txt`. This should work fine.

install.py
```diff
@@ -10,7 +10,19 @@
...
 import datetime
+import importlib
 
+def install_and_import(package):
+    import importlib
+    try:
+        importlib.import_module(package)
+    except ImportError:
+        import pip
+        pip.main(['install', package])
```
gavindsouza, line 21:
```suggestion
		subprocess.check_call('{0} -m pip install {1}'.format(sys.executable, package))
```
Start it as a subprocess instead. 

**References:**
1. https://pip.pypa.io/en/stable/user_guide/?highlight=_internal#using-pip-from-your-program
2. https://stackoverflow.com/a/47059613/10309266

install.py
```diff
@@ -95,14 +106,23 @@ def check_distribution_compatibility():
...
+    if platform.system() == 'Linux':
+        if int(platform.python_version_tuple()[0]
+               + platform.python_version_tuple()[1]) <= 3.7:
+            current_dist = platfor.dist()
+        else:
+            current_dist = \
+                distro.linux_distribution(full_distribution_name=True)
+
+            return (current_dist[0].lower(), current_dist[1].rsplit('.'
+                    )[0])
```
Alchez, line 121: You can use some other [`distro` functions](https://distro.readthedocs.io/en/latest/#consolidated-accessor-functions) for what you're trying to do:

```suggestion
            return (distro.id(), distro.major_version())
```

install.py
Alchez: Will this need to be done again in case there's another dependency to be added?

install.py
```diff
@@ -10,7 +10,18 @@
...
 import warnings
 import datetime
+import importlib
 
+def install_and_import(package):
+    import importlib
+    try:
+        importlib.import_module(package)
+    except ImportError:
+		subprocess.check_call('{0} -m pip install {1}'.format(sys.executable, package))
```
gavindsouza, line 20: I think `check_call` takes in an iterable for a command, like 
```python
>>> subprocess.check_call([sys.executable, "-m", "pip", "install", package]) 
```

install.py
```diff
@@ -10,7 +10,18 @@
 import platform
 import warnings
 import datetime
+import importlib
 
+def install_and_import(package):
+    import importlib
```
gavindsouza, line 16:
```suggestion
```

nikunj1222:
Hi guys it seems all checks are passed. Let me know if you need any more changes..

install.py
```diff
@@ -10,7 +10,18 @@
 import platform
 import warnings
 import datetime
+import importlib
 
+def install_and_import(package):
```
gavindsouza, line 15: * Rename it to `install_python_package` instead, that'd fit with the convention and is more intuitive. 
* Move this function down while maintaining 2 line spacing with the other functions

install.py
```diff
@@ -10,7 +10,18 @@
...
+    try:
+        importlib.import_module(package)
+    except ImportError:
+	print("Trying to Install required module: " + package)
+	subprocess.check_call([sys.executable, "-m", "pip", "install", package])
+    finally:
+        globals()[package] = importlib.import_module(package)
+
+install_and_import('distro')
```
gavindsouza, line 24: This should be called as the last step in the `install_prerequisites` function

install.py
```diff
@@ -95,14 +105,16 @@ def check_distribution_compatibility():
...
+
+    # return distribution name and major version
+
+	if platform.system() == 'Linux':
+            return (distro.id(), distro.major_version())
+	
+        elif platform.system() == 'Darwin':
+
+            current_dist = platform.mac_ver()
+            return ('macos', current_dist[0].rsplit('.', 1)[0])
```
gavindsouza, line 117: Indentations are off here. Use tabs only and remove unnecessary new-lines on L108, L110 and L115

install.py
```diff
@@ -10,7 +10,18 @@
...
+import importlib
 
+def install_and_import(package):
+    try:
+        importlib.import_module(package)
+    except ImportError:
+	print("Trying to Install required module: " + package)
+	subprocess.check_call([sys.executable, "-m", "pip", "install", package])
+    finally:
+        globals()[package] = importlib.import_module(package)
```
gavindsouza, line 22: Inconsistent indentation!

install.py
```diff
@@ -433,6 +445,7 @@ def parse_commandline_args():
 	with warnings.catch_warnings():
 		warnings.simplefilter("ignore")
 		setup_log_stream(args)
+		install_python_package('distro')
```
gavindsouza, line 448: this breaks if pip isn't already installed on the system. Specifically for Python versions <= PY2.7.9 and <= PY3.4

pip is installed later on in `install_prerequisites` as of right now.

nikunj1222: @gavindsouza  How shall this be resolved then. 
we will not install the prerequisite if the system is not compatible. OR we install prerequisite before checking the OS.

gavindsouza: By installing pip first. I've pushed the changes to this branch.

gavindsouza:
@nikunj1222 sorry we couldn't get this fixed earlier. but https://github.com/frappe/bench/pull/1048 has been merged. Thanks for your contribution.

PR #1024, opened 2020-06-11
fix: use remote_addr instead of user-defined header ip

Thunderbottom:
`$proxy_add_x_forwarded_for` sets the user defined ip in the request. changing it to `$remote_addr` ensures that nginx sets the remote ip itself, disregarding user-specified ip from the request.


revant:
We need this in Docker nginx template as well

PR #1030, opened 2020-07-01
fix: run backups every 6 hours, rather than every minute per 6 hours

Alchez:
What type of a PR is this?

- [ ] New Feature Submissions
- [x] Bug Fix

---

- [ ] Have you successfully run tests with your changes locally?

---

**Problem:**

For newer benches, the crontab would be set to generate backups every _minute_ per 6 hours, as opposed to running it every 6 hours.

**Solution:**

Fix the crontab generation accordingly

**Reference:**

From the [`python-crontab`](https://pypi.org/project/python-crontab/) package documentation:

> Proceeding Unit Confusion

> It is sometimes logical to think that job.hour.every(2) will set all proceeding units to ‘0’ and thus result in “0 */2 * * *”. Instead you are controlling only the hours units and the minute column is unaffected. The real result would be “* */2 * * *” and maybe unexpected to those unfamiliar with crontabs.

> There is a special ‘every’ method on a job to clear the job’s existing schedule and replace it with a simple single unit:

> ```
> job.every(4).hours()  == '0 */4 * * *'
> job.every().dom()     == '0 0 * * *'
> job.every().month()   == '0 0 0 * *'
> job.every(2).dows()   == '0 0 * * */2'
> ```

> This is a convenience method only, it does normal things with the existing api.

### DIFF
--- a/bench/patches/patches.txt
+++ b/bench/patches/patches.txt
@@ -5,3 +5,4 @@ bench.patches.v4.update_node
 bench.patches.v4.update_socketio
 bench.patches.v4.install_yarn #2
 bench.patches.v5.fix_user_permissions
+bench.patches.v5.fix_backup_cronjob
--- a/bench/patches/v5/fix_backup_cronjob.py
+++ b/bench/patches/v5/fix_backup_cronjob.py
@@ -1,0 +1,15 @@
+from bench.config.common_site_config import get_config
+from crontab import CronTab
+
+
+def execute(bench_path):
+	"""
+		This patch fixes a cron job that would backup sites every minute per 6 hours
+	"""
+
+	user = get_config(bench_path=bench_path).get('frappe_user')
+	user_crontab = CronTab(user=user)
+
+	for job in user_crontab.find_comment("bench auto backups set for every 6 hours"):
+		job.every(6).hours()
+		user_crontab.write()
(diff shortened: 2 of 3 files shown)


Alchez:
@gavindsouza, this might need a patch for existing systems, can you help with that please?

gavindsouza:
Sample cron record created by the buggy `job.hour.every(6)`:
```
* */6 * * * cd /absolute/path/to/bench && /Users/gavin/.pyenv/versions/3.7.7/bin/bench --verbose --site all backup  >> /absolute/path/to/bench/logs/backup.log 2>&1 #   bench auto backups set for every 6 hours
```

Parsing the crontab file, line by line, and replacing the cron expression if it matches the backup command is what seems like an acceptable solution to me. Will you write the patch for this?

Alchez:
@gavindsouza, I've added a patch for the cron fix, but I'm not sure how to run it. A `bench migrate` didn't do that.

gavindsouza:
@Alchez `bench migrate` is a frappe command. `bench update` runs the bench patches. You can run it with the `--patch` flag to run patches and migrate all sites on the current bench

Alchez:
@gavindsouza, I ran `bench update --patch` and even `bench update` but the patch doesn't seem to run. Can you test on your side and see if you can get it to trigger?

gavindsouza:
@Alchez works for me. Check `patches.txt` in your bench; It records which patches are run and need not be run again. Remove the record if it exists and try again.

Alchez:
@gavindsouza, my `patches.txt` in the root bench directory only has one patch in it - `bench.patches.v3.deprecate_old_config`

Unless you mean a different file, I can't seem to make the patch run. If you think it's good to go though, we can move this ahead.

PR #1038, opened 2020-07-24
feature: Added http2 to SSL site and push preload

fromthestone:
Hi dev folks,
I noticed that bench in production, using Nginx, it's not configured to use http2 when Frappe or ERPNext are served over SSL connection.

With a tiny modification, I have added `http2` features to nginx.conf configuration
I have also added `http2_push_preload on` to nginx.conf to preload the push of files to clients

adding http2 has these goals:

- Create a negotiation mechanism that allows clients and servers to elect to use HTTP 1.1, 2.0, or potentially other non-HTTP protocols.

- Maintain high-level compatibility with HTTP 1.1 (for example with methods, status codes, URIs, and most header fields).

- Decrease latency to improve page load speed in web browsers

- data compression of HTTP headers

- HTTP/2 Server Push

- pipelining of requests

- fixing the head-of-line blocking problem in HTTP 1.x

- multiplexing 

- multiple requests over a single TCP connection

Everything it's explained more in detail in these pages:

- [https://medium.com/@mena.meseha/http-2-server-push-tutorial-d8714154ef9a](https://medium.com/@mena.meseha/http-2-server-push-tutorial-d8714154ef9a)

- [https://en.wikipedia.org/wiki/HTTP/2](https://en.wikipedia.org/wiki/HTTP/2)

- [https://www.nginx.com/blog/nginx-1-13-9-http2-server-push](https://www.nginx.com/blog/nginx-1-13-9-http2-server-push)

I hope this feat request could be approved and it will be ported also to version 11 and 12

Waiting for your comments

### DIFF
--- a/bench/config/templates/nginx.conf
+++ b/bench/config/templates/nginx.conf
@@ -13,7 +13,7 @@ map {{ from_variable }} {{ to_variable }} {
 {%- macro server_block(bench_name, port, server_names, site_name, sites_path, ssl_certificate, ssl_certificate_key) %}
 server {
 	{% if ssl_certificate and ssl_certificate_key %}
-	listen {{ port }} ssl;
+	listen {{ port }} ssl http2;
 	{% else %}
 	listen {{ port }};
 	{% endif %}
@@ -25,6 +25,9 @@ server {
 		;
 
 	root {{ sites_path }};
+	{% if ssl_certificate and ssl_certificate_key %}
+	http2_push_preload on;
+	{% endif %}
 
 	{% if allow_rate_limiting %}
 	limit_conn per_host_{{ bench_name_hash }} 8;
@@ -60,6 +63,7 @@ server {
 	}
 
 	location /socket.io {
+
 		proxy_http_version 1.1;
 		proxy_set_header Upgrade $http_upgrade;
 		proxy_set_header Connection "upgrade";


fromthestone:
I currently use this conf on my ERPNext self hosted Instance.

I understand your concerns and I will post some screenshots as soon as possible

fromthestone:
> Are you sure this works?
> 
I would say that HTTP/2 improvement it speeds up ERPNext, and with some internal tests, it could be used also in erpnext.com commercial instances bringing speed and modernity to this commercial service.

@adityahase  **http/2 protocol is fully supported by Nginx and by all browsers from 3/4 years to today.**
Consider that Google itself already using version 3 of the HTTP protocol and major CDN like Cloudfront, KeyCDN, MaxCDN, and many others also using http/2 protocol for several years.

It works only with SSL enabled -> `{% if ssl_certificate and ssl_certificate_key %}

Furthermore enabling the `push preload` in Nginx, lets ERPNext preload and push `JS` and `CSS` assets faster.

My ERPNext instance already working in HTTP/2 for a year without any issue
image

You can also check how long it takes to load the desk and also check how assets are pushed ;-)
image



Other advantages:

- **Single Connection**. Only one connection to the server is used to load a website, and that connection remains open as long as the website is open. This reduces the number of round trips needed to set up multiple TCP connections.

- **Multiplexing**. Multiple requests are allowed at the same time, on the same connection. Previously, with HTTP/1.1, each transfer would have to wait for other transfers to complete.

- **Server Push**. Additional resources can be sent to a client for future use.

- **Prioritization**. Requests are assigned dependency levels that the server can use to deliver higher priority resources faster.

- **Binary**. Makes HTTP/2 easier for a server to parse, more compact and less error-prone. No additional time is wasted translating information from the text to binary, which is the computer’s native language.

- **Header Compression**. HTTP/2 uses HPACK compressions, which reduces overhead. Many headers were sent with the same values in every request in HTTP/1.1.


> Does this log you out?
> I mean when you do something like visit desk and then website?

Not at all, it does not interfere with cookies, it's only a benefit because will compress headers better and transport data in binary mode. HTTP/1.1 doesn't do all these things.


## CONS
The only cons that till today I found is the browser support.
All browsers before half of 2015 are doesn't support HTTP/2

here the list: [https://caniuse.com/#feat=http2](https://caniuse.com/#feat=http2)

## PRODUCTION USE CONSIDERATION
If the team and community want to test more http/2 use, we can add a switch in SSL and LetsEncrypt commands in order to enable http2 and push preload feature.

What do you think?
Waiting for community comments

fromthestone:
@adityahase I completely misunderstood.
Now I got what do you mean, sorry :-)
Normally at work, we don't use website but just only ERP, that's the reason because I never noticed this issue.
Let's do like this... close this PR and give me some time to fix the issue and I will open a new PR

PR #1039, opened 2020-07-31
feat: bench start --no-prefix

ABHISHEK-KEDAR-21:
Is it not awesome if anyone is not interested in viewing the process prefixed with the log on the terminal window.

By adding this option `--no-prefix` while `bench start` we can achieve it.

I feel this will help in finding any error or any log more easily and you can copy also from the terminal window efficiently without the process prefixed.

image


gavindsouza:
@ABHISHEK-KEDAR-21 only fixes accepted on the master branch! Closing this.

PR #1040, opened 2020-07-31
feat: bench start --no-prefix

ABHISHEK-KEDAR-21:
Is it not awesome if anyone is not interested in viewing the process prefixed with the log on the terminal window.

By adding this option `--no-prefix` while `bench start` we can achieve it.

I feel this will help in finding any error or any log more easily and you can copy also from the terminal window efficiently without the process prefixed.

image

### DIFF
--- a/bench/commands/utils.py
+++ b/bench/commands/utils.py
@@ -8,11 +8,12 @@ import click
 
 @click.command('start', help="Start Frappe development processes")
 @click.option('--no-dev', is_flag=True, default=False)
+@click.option('--no-prefix', is_flag=True, default=False)
 @click.option('--concurrency', '-c', type=str)
 @click.option('--procfile', '-p', type=str)
-def start(no_dev, concurrency, procfile):
+def start(no_dev, concurrency, procfile, no_prefix):
 	from bench.utils import start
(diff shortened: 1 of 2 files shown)


bench/commands/utils.py
```diff
@@ -8,11 +8,12 @@
 
 @click.command('start', help="Start Frappe development processes")
 @click.option('--no-dev', is_flag=True, default=False)
+@click.option('--no-prefix', is_flag=True, default=False)
```
gavindsouza, line 11: Can you add a description for this flag. Might be unclear what this is supposed to do from the CLI

ABHISHEK-KEDAR-21: sure

ABHISHEK-KEDAR-21: Where should I add the description? Can you please guide me.

gavindsouza: `@click.option` accepts a `help` parameter where you can add it.

ABHISHEK-KEDAR-21: I was assuming that only, just wanted to make sure.
Thanks @gavindsouza

ABHISHEK-KEDAR-21: Can you please check now. I have added.

PR #1048, opened 2020-08-14
fix: ubuntu 20.04 easy install hang half way

adamtang79:
fix: provide smooth easy install experience

### DIFF
--- a/bench/playbooks/roles/dns_caching/handlers/main.yml
+++ b/bench/playbooks/roles/dns_caching/handlers/main.yml
@@ -1,4 +1,4 @@
 ---
-- name: Restart network manager
+- name: restart network manager
   service: name=NetworkManager state=restarted
 ...
--- a/install.py
+++ b/install.py
@@ -74,11 +74,11 @@ def check_system_package_managers():
 			raise Exception('Cannot find any compatible package manager!')
 
 
-def check_distribution_compatibility():
-	dist_name, dist_version = get_distribution_info()
+def check_distribution_compatibility(args):
+	dist_name, dist_version = get_distribution_info(args)
 	supported_dists = {
 		'macos': [10.9, 10.10, 10.11, 10.12],
-		'ubuntu': [14, 15, 16, 18, 19],
+		'ubuntu': [14, 15, 16, 18, 19, 20],
 		'debian': [8, 9, 10],
 		'centos': [7]
 	}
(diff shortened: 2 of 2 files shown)


install.py
```diff
@@ -403,6 +428,16 @@ def parse_commandline_args():
...
 	parser.add_argument('--container', dest='container', default=False, action='store_true', help='Use if you\'re creating inside LXC')
+
+	# for detecting dist; backward compatible if specify 2
+
+	parser.add_argument(
+		'--python-version',
+		dest='python_version',
+		default='3',
+		help='For detecting dist. Backward compatible if specify 2. Default is 3.'
+	)
```
gavindsouza, line 439: Instead of asking for user input, we can check the Python version used for executing the script by `sys.version_info.major`

bench/playbooks/roles/dns_caching/handlers/main.yml
```diff
@@ -1,4 +1,4 @@
 ---
-- name: Restart network manager
+- name: restart network manager
```
gavindsouza, line 2: Why?

install.py
```diff
@@ -93,11 +93,36 @@ def check_distribution_compatibility():
...
 	if platform.system() == "Linux":
-		current_dist = platform.dist()
+		if args.python_version == "3":
+			install_package('pip3', 'python3-pip')
+
+			import_with_install('distro')
+
+			current_dist = distro.linux_distribution(full_distribution_name=True)
+		else:
+			current_dist = platform.dist()
```
gavindsouza, line 125: `platform.dist` is functional until `PY3.7` and is only deprecated in `PY3.8`. Check the minor version too, and install distro only if `PY3.8`+.

bench/playbooks/roles/dns_caching/handlers/main.yml
adamtang79: because below will cause the install process exit



- name: Add dnsmasq to network config
  lineinfile: >
    dest=/etc/NetworkManager/NetworkManager.conf
    regexp="dns="
    line="dns=dnsmasq"
    state=present
  when: result.stat.exists
  notify:
  - restart network manager

install.py
adamtang79: removed

install.py
adamtang79: using sys.version_info and only > 3.7 to install distro

bench/playbooks/roles/dns_caching/handlers/main.yml
gavindsouza: This shouldn't make any difference in your setup. It's merely for [documentation/reference purpose](https://docs.ansible.com/ansible/latest/reference_appendices/playbooks_keywords.html).

adamtang79: 1584859181.448698-32385133842846/AnsiballZ_lineinfile.py && sleep 0’
<127.0.0.1> EXEC /bin/sh -c ‘rm -f -r /root/.ansible/tmp/ansible-tmp-1584859181.448698-32385133842846/ > /dev/null 2>&1 && sleep 0’
ERROR! The requested handler ‘restart network manager’ was not found in either the main handlers list nor in the listening handlers list
File “install.py”, line 437, in

adamtang79: Above error gone after this fix.
This happened 12 months ago and no one fix it until today.

gavindsouza: yeah, I just checked and there was an issue with the title

gavindsouza:
LGTM

PR #1053, opened 2020-08-20
fix: warning on pip reinstall

adamtang79:
frappe version-12[-hotfix] has six==1.13.0


gavindsouza (changes requested):
how about we just unpin six instead? The warning disappears irrespective of Frappe's branch or dependency's version

PR #1062, opened 2020-09-03
fix: Don't show logs in stdout if command run outside bench directory

### LINKED ISSUE #1057: Double messages logged when commands executed outside a bench directory
Screenshot 2020-09-04 at 4 47 12 PM

gavindsouza:
Screenshot 2020-09-10 at 10 17 13 AM

if bench commands are executed under the bench directory, they get populated under `./logs/bench.log`. In case of commands executed outside the bench directories, this PR adds a `NullHandler` to the existing bench logger.

Fixes #1057

### DIFF
--- a/bench/cli.py
+++ b/bench/cli.py
@@ -27,7 +27,7 @@ def cli():
 	command = " ".join(sys.argv)
 
 	change_working_directory()
-	logger = setup_logging() or logging.getLogger(bench.PROJECT_NAME)
+	logger = setup_logging()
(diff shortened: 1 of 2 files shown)


saurabh6790:
@gavindsouza can we maintain only one log file?

gavindsouza:
@saurabh6790 I propose we maintain only `./logs/bench.log` and not log bench operations executed outside any bench directory

adityahase (approved):

PR #1070, opened 2020-09-07
fix: Update Letsencrypt Renew Cron Pattern

abhishekbalam:
- [x] Changed from `* * 1 * *` (At every minute on day-of-month 1)  to `0 0 */1 * *` (At 00:00 on every day-of-month).
- [ ] Add patch to fix existing set ups


adityahase (approved):

bench/patches/v5/fix_letsencrypt_renew_frequency.py
```diff
@@ -0,0 +1,19 @@
+from bench.config.common_site_config import get_config
+from crontab import CronTab
+
+
+def execute(bench_path):
+	"""
+		This patch fixes a cron job that would renew letsencrypt certificate
+	"""
+
+    job_command = '/opt/certbot-auto renew --force-renewal -a nginx --post-hook "systemctl reload nginx"'
```
adityahase, line 10: Indentation?

abhishekbalam: Fixing

bench/patches/v5/fix_letsencrypt_renew_frequency.py
```diff
@@ -0,0 +1,19 @@
...
+	system_crontab = CronTab(user='root')
```
gavindsouza, line 11: Does this mean we now need sudo to create a new bench?

abhishekbalam: No, this runs only if the letsencrypt command entry exists in cron

bench/patches/v5/fix_letsencrypt_renew_frequency.py
```diff
@@ -0,0 +1,19 @@
...
+	job_command = '/opt/certbot-auto renew -a nginx --post-hook "systemctl reload nginx"'
+	system_crontab = CronTab(user='root')
+	job_comment = "Renew lets-encrypt every month"
+
+	for job in system_crontab.find_comment(job_comment):
+		system_crontab.remove(job)
+		job = system_crontab.new(command=job_command, comment=job_comment)
+		job.setall('0 0 */1 * *') # Run at 00:00 every day-of-month
+		system_crontab.write()
+		break
```
gavindsouza, line 19: why `break`?

gavindsouza (commented):
patch breaks during init operation as frappe user

abhishekbalam:
> patch breaks during init operation as frappe user

Yes, and it can't be avoided as the letsencrypt commands needs root access.
So changing the setup function logic to check for older entries and replacing them as the setup command needs sudo anyway.

gavindsouza:
@Mergifyio backport master

gavindsouza (approved):

PR #1079, opened 2020-10-03
fix(install): Allow override of frappe_branch and erpnext_branch

ceefour:
Allow override of frappe_branch and erpnext_branch
Regardless of args.version (which always has a default set).
Additionally, print useful log of (non-credentials) variables used.

What type of a PR is this?

- [ ] New Feature Submissions
- [x] Bug Fix

---

Currently `--frappe-branch` and `--erpnext-branch` is always ignored because `--version` has a default (`12`).

**Important:** Before this fix, it's not possible at all to install `version-13-beta` because of `install.py: error: argument --version: invalid int value: '13-beta`.

After this fix, even if version is something else, frappe-branch and erpnext-branch can be set manually.

---

---

> Screenshots/GIFs

### DIFF
--- a/install.py
+++ b/install.py
@@ -260,7 +260,10 @@ def install_bench(args):
 	extra_vars.update(bench_name=bench_name)
 
 	# Will install ERPNext production setup by default
-	log("Initializing bench using bench_name=%s frappe_branch=%s erpnext_branch=%s ..." % (bench_name, frappe_branch, erpnext_branch))
+	if args.without_erpnext:
+		log("Initializing bench {bench_name}:\n\tFrappe Branch: {frappe_branch}\n\tERPNext will not be installed due to --without-erpnext".format(bench_name=bench_name, frappe_branch=frappe_branch))
+	else:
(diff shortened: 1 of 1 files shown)


install.py
```diff
@@ -261,6 +260,7 @@ def install_bench(args):
 	extra_vars.update(bench_name=bench_name)
 
 	# Will install ERPNext production setup by default
+	log("Initializing bench using bench_name=%s frappe_branch=%s erpnext_branch=%s ..." % (bench_name, frappe_branch, erpnext_branch))
```
gavindsouza, line 263:
```suggestion
	log("Initializing bench {0}:\n\tFrappe Branch: {1}\n\tERPNext Branch: {2}".format(bench_name, frappe_branch, erpnext_branch))
```

Can we have something like this instead? Also, check if `--without-erpnext` or `--without-bench-setup` are passed and show the appropriate statement log here

gavindsouza (changes requested):

install.py
gavindsouza: >  Also, check if --without-erpnext or --without-bench-setup are passed and show the appropriate statement log here

Address this too ^

ceefour:
@gavindsouza Updated with more conditional and clearer logging.

gavindsouza (approved):

PR #1080, opened 2020-10-03
fix(utils): Raise exception when generate_command_cache failed

ceefour:
What type of a PR is this?

- [ ] New Feature Submissions
- [x] Bug Fix

---

`generate_command_cache()` is used by cli's `get_frappe_commands()`. Previous behavior ignores errors during call of `/home/frappe/frappe-bench/env/bin/python -m frappe.utils.bench_helper get-frappe-commands`, which breaks following code and results in confusing error messages.

With this change:

* error in `get_cmd_output()` will be raised as exception after printing it out
* additional logging is provided, so `/home/frappe/frappe-bench/logs/bench.log` contain useful information for user to diagnose

---

- [ ] Have you lint your code locally prior to submission?
- [ ] Have you successfully run tests with your changes locally?
- [ ] Does your commit message have an explanation for your changes and why you'd like us to include them?

---

> Screenshots/GIFs

After this change, error message is clearer:

image

logs/bench.log also contain useful information:

image

### DIFF
--- a/bench/utils.py
+++ b/bench/utils.py
@@ -1061,7 +1061,7 @@ def generate_command_cache(bench_path='.'):
 		return json.loads(output)
 
 	except subprocess.CalledProcessError as e:
-		logger.error('generate_command_cache(\'%s\') failed executing: %s', bench_path, command, exc_info=e)
+		logger.error("generate_command_cache('%s') failed executing: %s", bench_path, command, exc_info=e)
 		if hasattr(e, "stderr"):
 			print(e.stderr.decode('utf-8'))
 		raise e
--- a/bench/utils.py
+++ b/bench/utils.py
@@ -1053,14 +1053,18 @@ def generate_command_cache(bench_path='.'):
 		os.remove(bench_cache_file)
 
 	try:
-		output = get_cmd_output("{0} -m frappe.utils.bench_helper get-frappe-commands".format(python), cwd=sites_path)
+		command = "{0} -m frappe.utils.bench_helper get-frappe-commands".format(python)
+		logger.debug('generate_command_cache(\'%s\') executing: %s', bench_path, command)
+		output = get_cmd_output(command, cwd=sites_path)
 		with open(bench_cache_file, 'w') as f:
 			json.dump(eval(output), f)
 		return json.loads(output)
 
 	except subprocess.CalledProcessError as e:
+		logger.error('generate_command_cache(\'%s\') failed executing: %s', bench_path, command, exc_info=e)
 		if hasattr(e, "stderr"):
 			print(e.stderr.decode('utf-8'))
+		raise e
 
 
 def clear_command_cache(bench_path='.'):
--- a/bench/utils.py
+++ b/bench/utils.py
@@ -1054,7 +1054,7 @@ def generate_command_cache(bench_path='.'):
 
 	try:
 		command = "{0} -m frappe.utils.bench_helper get-frappe-commands".format(python)
-		logger.debug('generate_command_cache(\'%s\') executing: %s', bench_path, command)
+		logger.debug("generate_command_cache('%s') executing: %s", bench_path, command)
 		output = get_cmd_output(command, cwd=sites_path)
 		with open(bench_cache_file, 'w') as f:
 			json.dump(eval(output), f)


bench/utils.py
```diff
@@ -1053,14 +1053,18 @@ def generate_command_cache(bench_path='.'):
...
+		output = get_cmd_output(command, cwd=sites_path)
 		with open(bench_cache_file, 'w') as f:
 			json.dump(eval(output), f)
 		return json.loads(output)
 
 	except subprocess.CalledProcessError as e:
+		logger.error('generate_command_cache(\'%s\') failed executing: %s', bench_path, command, exc_info=e)
 		if hasattr(e, "stderr"):
 			print(e.stderr.decode('utf-8'))
+		raise e
```
gavindsouza, line 1067: What is the reason for raising the error?

bench/utils.py
```diff
@@ -1053,14 +1053,18 @@ def generate_command_cache(bench_path='.'):
 		os.remove(bench_cache_file)
 
 	try:
-		output = get_cmd_output("{0} -m frappe.utils.bench_helper get-frappe-commands".format(python), cwd=sites_path)
+		command = "{0} -m frappe.utils.bench_helper get-frappe-commands".format(python)
+		logger.debug('generate_command_cache(\'%s\') executing: %s', bench_path, command)
```
gavindsouza, line 1057:
```suggestion
		logger.debug("generate_command_cache('%s') executing: %s", bench_path, command)
```

bench/utils.py
```diff
@@ -1053,14 +1053,18 @@ def generate_command_cache(bench_path='.'):
...
+		output = get_cmd_output(command, cwd=sites_path)
 		with open(bench_cache_file, 'w') as f:
 			json.dump(eval(output), f)
 		return json.loads(output)
 
 	except subprocess.CalledProcessError as e:
+		logger.error('generate_command_cache(\'%s\') failed executing: %s', bench_path, command, exc_info=e)
```
gavindsouza, line 1064:
```suggestion
		logger.error("generate_command_cache('%s') failed executing: %s", bench_path, command, exc_info=e)
```

gavindsouza (changes requested):
Can you add a screenshot as to what is currently happening and what problem are you trying to fix?

bench/utils.py
ceefour: generate_command_cache() is used by cli's get_frappe_commands(). Previous behavior ignores errors during call of /home/frappe/frappe-bench/env/bin/python -m frappe.utils.bench_helper get-frappe-commands, which breaks following code and results in confusing error messages.

By not throwing exception (previous behavior), execution will continue but will fail anyway later, with even more confusing/unpredictable error message that is hard to diagnose.

ceefour:
I have added screenshot in description.

Before this PR, behavior will be unpredictable depending on exact bench command that is being run. Which is precisely the point of this PR.

Usually the only clue of error was just "No module named frappe" followed by command-specific error message which masks the real issue, which makes diagnostic difficult.

ceefour:
@gavindsouza Here's an example scenario: https://about.lovia.life/docs/infrastructure/erpnext/install-erpnext-on-ubuntu-vm/#troubleshooting-3-psycopg2

So in case we run command `bench get-app erpnext ...` but installation of "frappe" app failed, e.g. due to low disk space, we'll get very confusing error message:

`TypeError: unsupported operand type(s) for +: 'NoneType' and 'list'`

It's even more confusing when it happened inside of Ansible / easy install script.

That's why throwing the error is necessary to fail earlier.

bench/utils.py
gavindsouza: Hey @ceefour, we want failures under `generate_command_cache` to be silent and non-blocking. 

So, I believe we shouldn't raise here, rather fix `get_frappe_commands` in `cli.py` to make sure it always returns a list, empty or filled. In that case, the screenshot with the error you shared earlier wouldn't arise.

Sorry for the delay in response.

gavindsouza (changes requested):

gavindsouza:
@ceefour Any update on this?

gavindsouza:
The issues this PR tried to fix have been addressed by https://github.com/frappe/bench/pull/1197. Closing this

ceefour:
Thank youuuu @gavindsouza

Issue #1082, opened 2020-10-03
bench init should exit with status 1 on error during "Installing frappe"


ceefour:
Issue: Bug report

Please make sure your issue is reproducible on the latest bench version. The currently supported branches are:
 - PyPI [latest](https://pypi.org/project/frappe-bench/) 
 - master         (minor bug fixes)
 - v5.x           (Merged with develop on every release)
 - develop        (all updates)

**Do the checklist before filing an issue:**
 - [x] Can you replicate the issue on the supported bench versions?
 - [x] Is this something you can debug and fix? Send a pull request! Bug fixes and documentation fixes are welcome

**Describe the bug** :chart_with_downwards_trend:
bench init should exit with status 1 on error during "Installing frappe"

**To Reproduce** :page_with_curl:
1. Ensure pg_config is not installed
2. Run: `bench init /home/frappe/frappe-bench --frappe-path https://github.com/frappe/frappe --frappe-branch version-13-beta --python python3`
3. Installing frappe had errors, but bench init continues, resulting in broken installation and **misleading** message **SUCCESS: Bench /home/frappe/frappe-bench initialized**

```
frappe@ip-172-30-0-218:~$ bench init /home/frappe/frappe-bench --frappe-path https://github.com/frappe/frappe --frappe-branch version-13-beta --python python3
WARN: bench is installed in editable mode!

This is not the recommended mode of installation for production. Instead, install the package from PyPI with: `pip install frappe-bench`

get_frappe_commands returned []
$ /usr/local/bin/virtualenv -q env -p python3

Getting frappe
$ git clone https://github.com/frappe/frappe --branch version-13-beta --depth 1 --origin upstream
Cloning into 'frappe'...
remote: Enumerating objects: 2924, done.
remote: Counting objects: 100% (2924/2924), done.
remote: Compressing objects: 100% (2724/2724), done.
remote: Total 2924 (delta 385), reused 1083 (delta 131), pack-reused 0
Receiving objects: 100% (2924/2924), 11.96 MiB | 2.01 MiB/s, done.
Resolving deltas: 100% (385/385), done.

Installing frappe
$ /home/frappe/frappe-bench/env/bin/pip install -q -U -e /home/frappe/frappe-bench/apps/frappe 
    ERROR: Command errored out with exit status 1:
[...]

PR #1083, opened 2020-10-03
fix: Exit bench init early if Frappe or other apps installed incorrectly

ceefour:
Fixes https://github.com/frappe/bench/issues/1082

What type of a PR is this?

- [x] Bug Fix

---

Detect if command failed, then exit with status 1.

In `bench/app.py` `install_app()` -> `bench.utils.exec_cmd()`

---

---

> Screenshots/GIFs

### DIFF
--- a/bench/app.py
+++ b/bench/app.py
@@ -174,8 +174,11 @@ def install_app(app, bench_path=".", verbose=False, no_cache=False, restart_benc
 	app_path = os.path.join(bench_path, "apps", app)
 	cache_flag = "--no-cache-dir" if no_cache else ""
 
-	pip_install_code = exec_cmd("{pip} install {quiet} -U -e {app} {no_cache}".format(
-		pip=pip_path, quiet=quiet_flag, app=app_path, no_cache=cache_flag))
+		pip_install_code = exec_cmd(
+		"{pip} install {quiet} -U -e {app} {no_cache}".format(
+			pip=pip_path, quiet=quiet_flag, app=app_path, no_cache=cache_flag
+		)
+	)
 	if pip_install_code:
 		sys.exit(pip_install_code)
 
--- a/bench/app.py
+++ b/bench/app.py
@@ -174,7 +174,10 @@ def install_app(app, bench_path=".", verbose=False, no_cache=False, restart_benc
 	app_path = os.path.join(bench_path, "apps", app)
 	cache_flag = "--no-cache-dir" if no_cache else ""
 
-	exec_cmd("{pip} install {quiet} -U -e {app} {no_cache}".format(pip=pip_path, quiet=quiet_flag, app=app_path, no_cache=cache_flag))
+	pip_install_code = exec_cmd("{pip} install {quiet} -U -e {app} {no_cache}".format(
+		pip=pip_path, quiet=quiet_flag, app=app_path, no_cache=cache_flag))
+	if pip_install_code:
+		sys.exit(pip_install_code)
 
 	if os.path.exists(os.path.join(app_path, 'package.json')):
 		exec_cmd("yarn install", cwd=app_path)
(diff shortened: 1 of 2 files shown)


bench/utils.py
```diff
@@ -301,7 +301,9 @@ def setup_app(app):
 
 
 def exec_cmd(cmd, cwd='.'):
-	import shlex
+	"""Executes 'cmd' as subprocess, then returns the return code.
+	If subprocess is successful, the return code is 0."""
+  import shlex
```
gavindsouza, line 306:
```suggestion
	"""Executes cmd with cwd via subprocess.call and returns the process exit code"""
	import shlex
```

Give more info with lesser words. Fix indentation

bench/app.py
```diff
@@ -174,7 +174,10 @@ def install_app(app, bench_path=".", verbose=False, no_cache=False, restart_benc
 	app_path = os.path.join(bench_path, "apps", app)
 	cache_flag = "--no-cache-dir" if no_cache else ""
 
-	exec_cmd("{pip} install {quiet} -U -e {app} {no_cache}".format(pip=pip_path, quiet=quiet_flag, app=app_path, no_cache=cache_flag))
+	pip_install_code = exec_cmd("{pip} install {quiet} -U -e {app} {no_cache}".format(
+		pip=pip_path, quiet=quiet_flag, app=app_path, no_cache=cache_flag))
```
gavindsouza, line 178:
```suggestion
		pip_install_code = exec_cmd(
		"{pip} install {quiet} -U -e {app} {no_cache}".format(
			pip=pip_path, quiet=quiet_flag, app=app_path, no_cache=cache_flag
		)
	)
```
Styling fixes

bench/app.py
```diff
@@ -174,7 +174,10 @@ def install_app(app, bench_path=".", verbose=False, no_cache=False, restart_benc
...
+	if pip_install_code:
+		sys.exit(pip_install_code)
```
gavindsouza, line 180: Can you add a screenshot for this during execution?

gavindsouza (changes requested):

ceefour:
@gavindsouza About screenshot, I documented a lot of the diagnosis process here: https://about.lovia.life/docs/infrastructure/erpnext/install-erpnext-on-ubuntu-vm/#Troubleshooting_3_psycopg2

So this was what happened:

image

My patch stopped the execution if an error happened, since it doesn't make sense to continue....... which will throw another error anyway (with even more confusing error).

By stopping early, it's easier for user to figure out what went wrong, fix it, then they can retry.

gavindsouza:
@ceefour I get that part. I wasn't able to replicate failing dependency setups and I'd like to see how bench would exit with the proposed fix. Can you show me a replicable screenshot so that there are no unknowns before we go ahead with this change?

ceefour:
One way that happened to me is low disk space..

So to reproduce it:

1. Do not install any apps yet
2. Make it so the disk space is left only a bit (probably easier to do inside a Docker container) 
3. Do bench get-app erpnext.. When it fails installing frappe, you'll see it continue etc.
4. Compare behavior of above but with patch applied

bench/app.py
```diff
@@ -174,7 +174,13 @@ def install_app(app, bench_path=".", verbose=False, no_cache=False, restart_benc
 	app_path = os.path.join(bench_path, "apps", app)
 	cache_flag = "--no-cache-dir" if no_cache else ""
 
-	exec_cmd("{pip} install {quiet} -U -e {app} {no_cache}".format(pip=pip_path, quiet=quiet_flag, app=app_path, no_cache=cache_flag))
+		pip_install_code = exec_cmd(
```
gavindsouza, line 177: @ceefour Indentation error

gavindsouza (changes requested):

gavindsouza:
Now, bench exits whenever any `exec_cmd` fails via https://github.com/gavindsouza/bench/commit/3995b9237b07ef5bef2fbd6a35000b7faa88ab2c. That should fix the issue this PR was addressing.

ceefour:
Thank youuu @gavindsouza

PR #1088, opened 2020-10-07
fix: Show colours only if terminal exists and it supports colour

gavindsouza:


gavindsouza:
Made obsolete via #1208

PR #1108, opened 2021-01-25
fix: nginx proxy buffer

### LINKED ISSUE #1068: NGINX: upstream sent too big header
With `frappe-bench==5.2.1`, on a. fresh ubuntu 18.04 server, I get two errors (maybe are connected to each other).

### 1. NGINX_ upstream sent too big header

The website of my instance shows maintenance mode while desk is still available. Nginx logs the following error (`/var/log/nginx/error.log`):

```
[error]
  12957#12957: *842 upstream sent too big header while reading response header from upstream,
  client:[CLIENT_IP],
  server: site1.local,
  request: "GET /index HTTP/1.1",
  upstream: "http://127.0.0.1:8000/index",
  host: "[SERVER _IP]"
```

### 2. Fail2ban blocks my IP address

Maybe because of the failed website requests from above error?

## Solution

I was able to solve the first error by adding the following to `frappe-bench/config/nginx.conf` ([source](https://ma.ttias.be/nginx-proxy-upstream-sent-big-header-reading-response-header-upstream)):

```
server {
    # ...
    proxy_buffer_size          128k;
    proxy_buffers              4 256k;
    proxy_busy_buffers_size    256k;
}
```

Now fail2ban has no reason to block me anymore.

barredterra:
Close #1068

### DIFF
--- a/bench/config/templates/nginx.conf
+++ b/bench/config/templates/nginx.conf
@@ -29,6 +29,10 @@ server {
 	{% if allow_rate_limiting %}
 	limit_conn per_host_{{ bench_name_hash }} 8;
 	{% endif %}
+	
+	proxy_buffer_size 128k;
+	proxy_buffers 4 256k;
+	proxy_busy_buffers_size 256k;
 
 	{% if ssl_certificate and ssl_certificate_key %}
 	ssl on;


gavindsouza (approved):

barredterra:
@gavindsouza, I just ran into this problem again. Could we get a patch release that includes this fix?

PR #1124, opened 2021-02-22
perf: Faster CLI

adityahase:
This doesn't look tasteful, but at least it's fast. Well, I'll let you decide if it's fast. This is how long a typical bench CLI invocation takes.

```
$ time bench --site speed.com execute frappe.ping  
0.69s user 0.09s system 100% cpu 0.770 total
```
We can improve this by not importing code that isn't executed e.g. Jinja, Requests, etc, and shed ~300ms.
```
$ time bench --site speed.com execute frappe.ping  
0.28s user 0.06s system 101% cpu 0.329 total
```
### Profiling
Python provides a non-invasive way to profile import times with [`-X importtime`](https://docs.python.org/3/using/cmdline.html#id5). Executing `bench` doesn't seem to be compatible with this flag but, we can create `main.py`

```python3
from bench.cli import cli

if __name__ == "__main__":
	cli()
```
and execute
```
python -X importtime ../../bench/bench/main.py --site speed.com execute frappe.ping
```
**Note:** `../../bench/bench/main.py` is the path where I had placed the main.py file.

```
import time: self [us] | cumulative | imported package
import time:       102 |        102 | zipimport
import time:       636 |        636 | _frozen_importlib_external
...
...
import time:       382 |        382 |     bench.commands.install
import time:       402 |       5915 |   bench.commands
import time:       308 |      36098 | bench.cli
"pong"
```
We can then use [tuna](https://github.com/nschloe/tuna) for visualization with something like.
```
python -X importtime ../../bench/bench/main.py --site speed.com execute frappe.ping 2> import.log
tuna import.log
```
### Screenshots
**Before**
Screenshot_2021-02-19 tuna - import log(15)
**After**
Screenshot_2021-02-22 tuna - import log

Actually, the CLI is slower than shown here, https://github.com/frappe/frappe/pull/12447 has a similar effect on Frappe.
Before Both PRs
```
$ time bench --site speed.com execute frappe.ping  
1.03s user 0.13s system 100% cpu 1.152 total
```
After this PR (and before frappe PR)
```
$ time bench --site speed.com execute frappe.ping  
0.69s user 0.09s system 100% cpu 0.770 total
```
After both PRs
```
$ time bench --site speed.com execute frappe.ping  
0.28s user 0.06s system 101% cpu 0.329 total
```
**Note:** If you're using editable install then you might see worse numbers (~200ms slower), if not then you might have to reinstall `bench` after every change.

### DIFF
--- a/bench/app.py
+++ b/bench/app.py
@@ -13,7 +13,6 @@ import sys
 # imports - third party imports
 import click
 import git
-import requests
 import semantic_version
 from six.moves import reload_module
 
@@ -91,6 +90,8 @@ def remove_from_excluded_apps_txt(app, bench_path='.'):
 		return write_excluded_apps_txt(apps, bench_path=bench_path)
 
 def get_app(git_url, branch=None, bench_path='.', skip_assets=False, verbose=False, restart_bench=True, overwrite=False):
+	import requests
+
 	if not os.path.exists(git_url):
 		if not is_git_url(git_url):
 			orgs = ['frappe', 'erpnext']
@@ -433,6 +434,8 @@ def install_apps_from_path(path, bench_path='.'):
 		get_app(app['url'], branch=app.get('branch'), bench_path=bench_path, skip_assets=True)
 
 def get_apps_json(path):
+	import requests
+
 	if path.startswith('http'):
 		r = requests.get(path)
 		return r.json()
--- a/bench/app.py
+++ b/bench/app.py
@@ -17,7 +17,6 @@ from six.moves import reload_module
 
 # imports - module imports
 import bench
-from bench.config.common_site_config import get_config
 from bench.utils import color, CommandFailedError, build_assets, check_git_for_shallow_clone, exec_cmd, get_cmd_output, get_frappe, restart_supervisor_processes, restart_systemd_processes, run_frappe_cmd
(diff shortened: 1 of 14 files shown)


Alchez:
@adityahase 

Just curious, are the benchmarks just for the `ping` function?

Wouldn't function imports lead to a slower call to the specific command, like `get-app` or `install-app`? What do the benchmarks look like for those commands?

gavindsouza (approved):

gavindsouza:
> Wouldn't function imports lead to a slower call to the specific command, like `get-app` or `install-app`? What do the benchmarks look like for those commands?

@Alchez I don't think these changes increase execution times for any particular command as the same modules would be imported in any way that are required by a certain command. So, best case scenario, 
`bench:cli` hits one of the functions this PR touches and only the required modules needs to be imported, which would make execution faster. On the other hand, for a function that requires all the modules to be imported won't suffer from degradation of performance as it would take the Python virtual machine the same amount of time to import the same set of modules in whatever order.

But I'm eager to see benchmarks, just out of curiosity. @adityahase do me let me know if I've missed or misunderstood anything here.

---

Anyhoo, merging this 👍🏼

PR #1125, opened 2021-02-25
fix: disable reloader in default Procfile template

ankush:
What type of a PR is this?

- [ ] New Feature Submissions
- [x] Bug Fix

---

Reloader causes connection to break during initial setup process. It should be disabled by default like Werkzeug.run_simple().

Discuss thread about this issue: https://discuss.erpnext.com/t/setup-failed-could-not-start-up-error-in-setup/36557/

---

---

> Screenshots/GIFs

### DIFF
--- a/bench/config/templates/Procfile
+++ b/bench/config/templates/Procfile
@@ -3,7 +3,10 @@ redis_cache: redis-server config/redis_cache.conf
 redis_socketio: redis-server config/redis_socketio.conf
 redis_queue: redis-server config/redis_queue.conf
 {% endif %}
-web: bench serve {% if webserver_port -%} --port {{ webserver_port }} {%- endif %}
+
+# Remove --noreload flag to enable reloading on python file changes.
+# Keep this flag until initial setup on site is finished.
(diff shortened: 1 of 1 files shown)


bench/config/templates/Procfile
```diff
@@ -3,7 +3,10 @@ redis_cache: redis-server config/redis_cache.conf
 redis_socketio: redis-server config/redis_socketio.conf
 redis_queue: redis-server config/redis_queue.conf
 {% endif %}
-web: bench serve {% if webserver_port -%} --port {{ webserver_port }} {%- endif %}
+
+# Remove --noreload flag to enable reloading on python file changes.
+# Keep this flag until initial setup on site is finished.
+web: bench serve {% if webserver_port -%} --port {{ webserver_port }} {%- endif %} --noreload
```
leela, line 9: 👍 

Is there a way to start bench in developer mode without touching procfile? can we use `bench start` with option?

ankush: This PR will likely make `--no-dev` command confusing. So we should fix that too.

Maybe we can have `--dev` command too that enables file reloader. This way we won't have to keep editing the procfile. 

@gavindsouza your thoughts ?

PR #1126, opened 2021-02-26
feat: skip redis config during supervisor setup

revant:
skips redis configuration during `bench setup supervisor --skip-redis`
In case redis is not installed locally and used as a remote service it causes redis processes to fail with command not found.

### DIFF
--- a/bench/commands/setup.py
+++ b/bench/commands/setup.py
@@ -42,9 +42,10 @@ def reload_nginx():
 @click.command("supervisor", help="Generate configuration for supervisor")
 @click.option("--user", help="optional user argument")
 @click.option("--yes", help="Yes to regeneration of supervisor config", is_flag=True, default=False)
-def setup_supervisor(user=None, yes=False):
+@click.option("--skip-redis", help="Skip redis configuration", is_flag=True, default=False)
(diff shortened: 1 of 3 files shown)


gavindsouza:
@revant LGTM. Can you update the [Documentation Page](https://frappeframework.com/docs/user/en/bench/bench-commands) too?

revant:
https://frappeframework.com/docs/user/en/bench/bench-commands#setup-commands

Setup commands don't document optional param here.
Should I add an option just for supervisor command?

gavindsouza:
Yeah, in a format similar to Frappe commands' docs. 

eg: [Site Deletion](https://frappeframework.com/docs/user/en/bench/frappe-commands#site-deletion) section with a seperate page for [usage reference](https://frappeframework.com/docs/user/en/bench/reference/drop-site)

revant:
added it somewhere else. I think it is relevant where supervisor is setup.

okay?

gavindsouza:
Yeah. Updating the reference guide is cool but IMO this should be on the documentation first.

revant:
If someone adds the setup section, I'll edit it.

gavindsouza (approved):

PR #1127, opened 2021-02-26
feat: Add progress bar for Easy Install Script

### LINKED ISSUE #752: Task progress bar (% completion) in the install.py script
Issue: Feature Request

Sometime some tasks take a long period of time and its hard to figure out if something has gone wrong and its stuck or its still working

It would be nice to have a progress bar (% completion display) specially for bigger tasks. 

**Additional context**
I am happy to contribute, if someone can guide a little

rtdany10:
Adds progress bar to easy install script so that user gets an idea if the installation is going on or if the system is stuck.
Closes #752

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,4 +7,5 @@ requests==2.22.0
 semantic-version==2.8.2
 setuptools
 six
+tqdm
 virtualenv
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,5 +7,4 @@ requests==2.22.0
 semantic-version==2.8.2
 setuptools
 six
-tqdm
 virtualenv
--- a/.travis.yml
+++ b/.travis.yml
@@ -81,7 +81,7 @@ matrix:
     script: sudo python $TRAVIS_BUILD_DIR/install.py --user travis --run-travis --production --verbose
 
 install:
-  - pip install urllib3 pyOpenSSL ndg-httpsclient pyasn1
+  - pip install urllib3 pyOpenSSL ndg-httpsclient pyasn1 tqdm
 
   - if [ $TEST == "bench" ];then
        wget -q -O /tmp/wkhtmltox.tar.xz https://github.com/frappe/wkhtmltopdf/raw/master/wkhtmltox-0.12.3_linux-generic-amd64.tar.xz;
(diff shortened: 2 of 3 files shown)


rtdany10:
Can't get to import tqdm into the virtual environment in Travis build.
Guidance/help to achieve the same appreciated.

requirements.txt
```diff
@@ -7,4 +7,5 @@ requests==2.22.0
 semantic-version==2.8.2
 setuptools
 six
+tqdm
```
gavindsouza, line 10: The `requirements.txt` file tracks `bench`'s dependencies, not the `install.py` script's.

Remove it from here. What you want to do is add `import_with_install('tqdm')` in the `setup_script_requirements` function which manages the script's dependencies.

https://github.com/frappe/bench/blob/f3809b00acc4bfa586e9a12116fb6bd262d3226e/install.py#L418-L421

gavindsouza (changes requested):

requirements.txt
rtdany10: Thank you, have made the changes. :)

gavindsouza (commented):
Travis is still failing. We can probably fix this but I'm unsure if this will be of much help. 90% of the execution time is taken up by 

https://github.com/frappe/bench/blob/f3809b00acc4bfa586e9a12116fb6bd262d3226e/install.py#L279

If you check the `site.yml` file, you'll notice that it's what does the actual setup. It seems that this PR won't be updating the progress bar for all updates in this single step either. And the progress bar might seem stuck for a very long time. You'd be better off running the script with the `--verbose` flag if you're worried about feedback from it.

rtdany10:
> Travis is still failing. We can probably fix this but I'm unsure if this will be of much help. 90% of the execution time is taken up by 
> 
> https://github.com/frappe/bench/blob/f3809b00acc4bfa586e9a12116fb6bd262d3226e/install.py#L279
> 
> If you check the `site.yml` file, you'll notice that it's what does the actual setup. It seems that this PR won't be updating the progress bar for all updates in this single step either. And the progress bar might seem stuck for a very long time. You'd be better off running the script with the `--verbose` flag if you're worried about feedback from it.

Just realized this. Maybe we can close this PR and the issue as well. The progress is just a gimmick!😅

PR #1129, opened 2021-03-02
fix: Merge v5.x into develop to get good version number

FHenry:
Into docker develop deployment get rig of the anoying message "upgrade bench to 5.3"


gavindsouza:
We don't need to do this. https://github.com/frappe/bench/pull/1138 is a better way to manage this. Thanks for raising this PR though 👍🏼

PR #1137, opened 2021-03-11
test: travis to GHA

hasnain2808:
From travis to GHA


gavindsouza:
Closing this PR considering the following reasons:

1. TravisCI works fine for us without any intervention. No reason to really switch to GHA.
2. Our GitHub actions' pipelines are usually choked by the Frappe and ERPNext repo's builds. It's painful enough to wait for over an hour to see my builds run there. If switching causes more of that, I'd rather not.
3. This PR seems to require some more work, and has been open for over 5 months now.

PR #1138, opened 2021-03-17
chore: update version

hasnain2808:
Bench shows that an update is available when on `develop` branch since the develop branch has a lower version
Solution: use a develop version to indicate this is develop branch


bench/utils.py
```diff
@@ -98,12 +98,13 @@ def check_latest_version():
 		return
 
 	if pypi_request.status_code == 200:
-		pypi_version_str = pypi_request.json().get('info').get('version')
-		pypi_version = Version(pypi_version_str)
-		local_version = Version(bench.VERSION)
+		if not bench.VERSION.endswith("dev"):
```
gavindsouza, line 101: How about you add the following even before the imports are made instead. 

```python
if bench.VERSION.endswith("dev"):
    return
```

I don't see the need to change anything else.

gavindsouza (changes requested):

gavindsouza (approved):

PR #1149, opened 2021-04-11
feat: use nice with backup cron job

sagarvora:
## Description

Over the time, the size of a site's database is bound to increase. When the backup command is being run in background, it consumes a lot of CPU. To give it a low priority, this PR proposes using `nice`.

**Patch tested locally.**

## Alternate solution

Use `nice` inside implementation of `bench backup` where the actual `dump` executables are being run. This way, it will also get less priority when the `bench backup` command is being run manually.


bench/utils.py
```diff
@@ -401,7 +401,11 @@ def setup_backups(bench_path='.'):
 	if bench.FRAPPE_VERSION == 4:
```
sagarvora, line 401: About time we remove support for version 4.

PR #1150, opened 2021-04-11
feat: soft restart (graceful reload to prevent downtime)

sagarvora:


bench/commands/utils.py
```diff
@@ -17,15 +17,23 @@ def start(no_dev, concurrency, procfile, no_prefix):
...
 	from bench.utils import restart_supervisor_processes, restart_systemd_processes
 	from bench.config.common_site_config import get_config
-	if get_config('.').get('restart_supervisor_on_update') or supervisor:
-		restart_supervisor_processes(bench_path='.', web_workers=web)
-	if get_config('.').get('restart_systemd_on_update') or systemd:
+
+	conf = get_config('.')
+
+	if not (hard or soft) and conf.get("default_to_soft_restart"):
+		soft = True
```
sagarvora, line 32: combine

bench/commands/utils.py
```diff
@@ -17,15 +17,21 @@ def start(no_dev, concurrency, procfile, no_prefix):
 
 
 @click.command('restart', help="Restart supervisor processes or systemd units")
+@click.option('--hard', is_flag=True, default=False)
+@click.option('--soft', is_flag=True, default=False)
```
ankush, line 21: Boolean flags might be a better fit here? Also, both defaulting to False seems a bit incorrect 🤔  The default should be current behaviour, right?

Ref: https://click.palletsprojects.com/en/7.x/options/#boolean-flags

sagarvora: Thanks for the review @ankush! Actually, the issue is that the default depends on a value set in `common_site_config.json` (`soft_restart`)

Maybe we can set the default to something like `None` that gets replaced with the config value inside the function.

gavindsouza: We can set `default=True` for the `--hard` option imo

Edit: Nope. That raises more issues. The current implementation seems okay

sagarvora:
Closing for now

PR #1152, opened 2021-04-14
fix: Replace certbot-auto with certbot

### LINKED ISSUE #1105: certbot-auto deprecated
Issue: Bug report

Ubuntu 20.04.1 LTS and Ubuntu 18.04.5 LTS

bench version
erpnext 13.x.x-develop
frappe 13.x.x-develop
I just did a bench update as well 

I followed the Guide on 
https://frappeframework.com/docs/user/en/bench/guides/lets-encrypt-ssl-setup

```
sudo -H bench setup lets-encrypt domain.com
Running this will stop the nginx service temporarily causing your sites to go offline
Do you want to continue? [y/N]: y
$ sudo systemctl stop nginx
$ /opt/certbot-auto  --config /etc/letsencrypt/configs/domain.com.cfg certonly
Skipping bootstrap because certbot-auto is deprecated on this system.
Your system is not supported by certbot-auto anymore.
Certbot cannot be installed.
Please visit https://certbot.eff.org/ to check for other alternatives.
nginx.conf already exists and this will overwrite it. Do you want to continue? [y/N]: n
```

**Possible Solution** :bookmark_tabs:
I followed this guide here and was able to install the cert. It does install certbot instead of certbot-auto but the command otherwise is the same.

abhishekbalam:
certbot-auto is depricated on many platforms. More information can be found [here](https://community.letsencrypt.org/t/certbot-auto-deprecated-explanation-and-solutions/139821)

This PR adds prerequisites for certbot and replaces all references to older commands.

Pls do not merge, testing is in progress.

Fixes #1105


abhishekbalam:
@revant @saurabh6790 

Removed setup scripts for certbot, it will be added in another PR later.

abhishekbalam:
Tested with @athul.

It works fine.

We need to update installation docs to ask users to install certbot seperately.

bench/config/templates/frappe_sudoers
```diff
@@ -15,6 +15,6 @@
 {{ user }} ALL = (root) NOPASSWD: {{ nginx }}
 {% endif %}
 
-{{ user }} ALL = (root) NOPASSWD: /opt/certbot-auto
+{{ user }} ALL = (root) NOPASSWD: /usr/bin/certbot
```
gavindsouza, line 18: If certbot exists in bin, it exists in PATH. If that is true, you should use `bench.utils.which` instead. If nothing is found, you can pass this path into the template after verification.

bench/config/lets_encrypt.py
```diff
@@ -105,20 +105,10 @@ def create_dir_if_missing(path):
...
-	from urllib.request import urlretrieve
-
-	certbot_path = get_certbot_path()
-	create_dir_if_missing(certbot_path)
-
-	if not os.path.isfile(certbot_path):
-		urlretrieve("https://dl.eff.org/certbot-auto", certbot_path)
-		os.chmod(certbot_path, 0o744)
-
-
```
gavindsouza, line 118: Check if certbot is installed and prompt the user to install it following some ABCs. Perhaps a link to a guide on Frappe Docs

bench/config/lets_encrypt.py
```diff
@@ -11,6 +11,7 @@
 from bench.config.production_setup import service
 from bench.config.site_config import get_domains, remove_domain, update_site_config
 from bench.utils import CommandFailedError, exec_cmd, update_common_site_config
+from distutils.spawn import find_executable
```
gavindsouza, line 14: Use `bench.utils.which` instead. We're trying to get rid of the distutils dependency.

bench/config/lets_encrypt.py
```diff
@@ -56,11 +57,10 @@ def create_config(site, custom_domain):
 
 def run_certbot_and_setup_ssl(site, custom_domain, bench_path, interactive=True):
 	service('nginx', 'stop')
-	get_certbot()
-
+	
 	try:
 		interactive = '' if interactive else '-n'
-		exec_cmd(f"{get_certbot_path()} {interactive} --config /etc/letsencrypt/configs/{custom_domain or site}.cfg certonly")
+		exec_cmd("{path} certonly {interactive} -c /etc/letsencrypt/configs/{site}.cfg".format(path=get_certbot_path(), interactive=interactive, site=custom_domain or site))
```
gavindsouza, line 63: Would be more readable as a f-string

bench/config/lets_encrypt.py
```diff
@@ -86,7 +86,7 @@ def run_certbot_and_setup_ssl(site, custom_domain, bench_path, interactive=True)
 def setup_crontab():
 	from crontab import CronTab
 
-	job_command = '/opt/certbot-auto renew -a nginx --post-hook "systemctl reload nginx"'
+	job_command = '{path} renew --nginx -q --renew-hook "systemctl reload nginx"'.format(path=get_certbot_path())
```
gavindsouza, line 89: Feel like this should check which service is being used, like `bench.utils.service` does

also, f-string

gavindsouza (changes requested):
You can use f-strings instead for everything here.

gavindsouza:
Any update on this? @abhishekbalam

gavindsouza:
Closing this due to inactivity (for ~7.5 months). Re-open when PR is ready.

PR #1155, opened 2021-04-22
feat: Verbosity control for setup requirements

gavindsouza:
Pip takes really long these days to resolve dependencies. It'd be nice to have some feedback and the option not to.


gavindsouza:
Not convinced that the default output should look like this:

Screenshot 2021-05-13 at 3 32 15 PM

~140 lines of output for when Frappe's dependencies are cached, and no hiccups while retrieving and installing them. This number is for Frappe alone and will significantly change if caches aren't present, also considering the number of apps installed on bench. We could be looking at >500 lines of pip output. 

Considering dynamic alternatives like a progressbar, or updating progress in place.

gavindsouza:
The dynamic feed feature _(switched on by adding `--use-feature='dynamic-feed'`)_  introduced by #1208 does something similar. Closing this since verbosity is fixed through the same PR too, to make pip verbose just pass `--verbose` after bench.

PR #1161, opened 2021-04-26
chore: Remove `set-default-site` command

NagariaHussain:
1. Remove `set-default-site` command. 
2. Update usage docs.

The `set-default-site` no longer works as expected and prints some output like (although `frappe` is installed correctly):

```
frappe app is not installed. Run the following command to install frappe
bench get-app https://github.com/frappe/frappe.git
...
```

`bench use <site-name>` should be used instead.

PR to update Frappe Docs: https://github.com/frappe/frappe_docs/pull/125

### DIFF
--- a/bench/commands/utils.py
+++ b/bench/commands/utils.py
@@ -98,12 +98,6 @@ def set_redis_socketio_host(host):
 	set_redis_socketio_host(host)
 
 
-@click.command('set-default-site', help="Set default site for bench")
-@click.argument('site')
-def set_default_site(site):
-	from bench.utils import set_default_site
-	set_default_site(site)
-
 
 @click.command('download-translations', help="Download latest translations")
 def download_translations():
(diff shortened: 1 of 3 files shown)


bench/commands/utils.py
```diff
@@ -98,7 +98,7 @@ def set_redis_socketio_host(host):
 	set_redis_socketio_host(host)
 
 
-@click.command('set-default-site', help="Set default site for bench")
+@click.command('set-default-site [DEPRECATED]', help="Set default site for bench")
```
gavindsouza: How about we just remove it completely?

Alternatively, we can execute `bench use` if this is executed. Although I'm leaning towards complete elimination.

gavindsouza (changes requested):

bench/commands/utils.py
gavindsouza: Also, update frappe docs along with the in-repo docs too.

NagariaHussain: > How about we just remove it completely?
> 
> 
> 
> Alternatively, we can execute `bench use` if this is executed. Although I'm leaning towards complete elimination. 

Ok.

gavindsouza (approved):

PR #1130, opened 2021-03-04
feat: Keep repos shallow if --reset and shallow_clone are specified in bench update --pull

casesolved-co-uk:
What type of a PR is this?

- [x] Changes to Existing Features

Saves disk space on production machines

---

- [x] Have you followed the guidelines in our Contributing document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](../pulls) for the same update/change?
- [x] Have you successfully run tests with your changes locally?
- [x] Does your commit message have an explanation for your changes and why you'd like us to include them?

---

> $ bench update --apps frappe --reset --pull
```
remote: Total 0 (delta 0), reused 0 (delta 0), pack-reused 0
From https://github.com/frappe/frappe
 * branch            develop    -> FETCH_HEAD
Backing up sites...
...
$ git fetch --depth=1 --no-tags upstream develop
remote: Total 0 (delta 0), reused 0 (delta 0), pack-reused 0
From https://github.com/frappe/frappe
 * branch            develop    -> FETCH_HEAD
$ git reset --hard upstream/develop
HEAD is now at a342ca2 fix: Get file name from path (#12437)
$ git reflog expire --all
$ git gc --prune=all
Counting objects: 3285, done.
Compressing objects: 100% (2627/2627), done.
Writing objects: 100% (3285/3285), done.
Total 3285 (delta 586), reused 3285 (delta 586)
$ find . -name "*.pyc" -delete

```

PR #1168, opened 2021-05-10
fix: git unrelated histories

ankush:
What type of a PR is this?

- [ ] New Feature Submissions
- [X] Bug Fix

---

Problem:

After updating bench to the latest version few users started getting "refusing to merge unrelated histories"

image

Root cause:
A recent change to allow keeping repo shallow after the update is fetching with `depth=1`, this will make `FETCH_HEAD` have just 1 latest commit, so git refuses to merge it with a previous full tree.

E.g. if a user does `bench update` it will trigger two git commands:
1. `git fetch --depth=1 upstream branch` (to check version update)
2. `git pull upstream branch`

here `git pull` tries to merge using shallow fetch and fails. 

image

reference: https://github.com/frappe/bench/pull/1130/files#diff-3722f4ad716221f5c6d162bf82efd37d96c6367678903aba4e51725dd57ee75bR353

Solution:
- Firstly, this should fail the update process and not proceed silently which can have unexpected side effects.
- don't use `depth=1` while fetching branch for checking version check.

### DIFF
--- a/bench/app.py
+++ b/bench/app.py
@@ -287,7 +287,7 @@ Here are your choices:
 					exec_cmd(reset_cmd, cwd=app_dir)
 			else:
 				exec_cmd("git pull {rebase} {remote} {branch}".format(rebase=rebase,
-					remote=remote, branch=branch), cwd=app_dir)
+					remote=remote, branch=branch), cwd=app_dir, fail_on_non_zero_return=True)
 			exec_cmd('find . -name "*.pyc" -delete', cwd=app_dir)
 
 
--- a/bench/app.py
+++ b/bench/app.py
@@ -277,7 +277,7 @@ Here are your choices:
 			if reset:
 				reset_cmd = "git reset --hard {remote}/{branch}".format(remote=remote, branch=branch)
 				if get_config(bench_path).get('shallow_clone'):
-					exec_cmd("git fetch --depth=1 --no-tags {remote} {branch}".format(remote=remote, branch=branch),
+					exec_cmd("git fetch {remote} {branch}".format(remote=remote, branch=branch),
 						cwd=app_dir)
 					exec_cmd(reset_cmd, cwd=app_dir)
 					exec_cmd("git reflog expire --all", cwd=app_dir)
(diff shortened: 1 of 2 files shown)


bench/utils.py
```diff
@@ -300,14 +300,19 @@ def setup_app(app):
 		setup_app(app)
 
 
-def exec_cmd(cmd, cwd='.'):
+def exec_cmd(cmd, cwd='.', fail_on_non_zero_return=False):
```
gavindsouza, line 303: `fail_on_non_zero_return` is wordy, yet fails to explain what actually would happen. If an exception would be raised, maybe rename this to `throw` or `raise`?

bench/utils.py
```diff
@@ -300,14 +300,19 @@ def setup_app(app):
...
 	import shlex
 	print("{0}$ {1}{2}".format(color.silver, cmd, color.nc))
 	cwd_info = "cd {0} && ".format(cwd) if cwd != "." else ""
 	cmd_log = "{0}{1}".format(cwd_info, cmd)
 	logger.debug(cmd_log)
 	cmd = shlex.split(cmd)
-	return_code = subprocess.call(cmd, cwd=cwd, universal_newlines=True)
+
+	if fail_on_non_zero_return:
+		return_code = subprocess.check_call(cmd, cwd=cwd, universal_newlines=True)
```
gavindsouza, line 312: On failure, will this display the `stderr` by the command or end with a Python traceback?

bench/app.py
```diff
@@ -277,7 +277,7 @@ def pull_apps(apps=None, bench_path='.', reset=False):
 			if reset:
 				reset_cmd = "git reset --hard {remote}/{branch}".format(remote=remote, branch=branch)
 				if get_config(bench_path).get('shallow_clone'):
-					exec_cmd("git fetch --depth=1 --no-tags {remote} {branch}".format(remote=remote, branch=branch),
+					exec_cmd("git fetch {remote} {branch}".format(remote=remote, branch=branch),
```
gavindsouza, line 280: This change is a bit unclear. Overall, I'm not certain what these changes will change. Can you add what is the expected behaviour of these changes to clear this up?

gavindsouza (changes requested):

PR #1182, opened 2021-06-08
feat: Changed function download_translations_p in bench/utils.py to use multithreading.

Aradhya-Tripathi:
the function uses multiple processes for an I/O bound task using threading would improve the performance significantly as spinning up processes takes longer than using threads.

### DIFF
--- a/bench/utils.py
+++ b/bench/utils.py
@@ -822,15 +822,15 @@ def update_translations_p(args):
 
 
 def download_translations_p():
-	import multiprocessing
+	from concurrent.futures import ThreadPoolExecutor
 
-	pool = multiprocessing.Pool(multiprocessing.cpu_count())
+	executor = ThreadPoolExecutor()
 
 	langs = get_langs()
 	apps = ('frappe', 'erpnext')
 	args = list(itertools.product(apps, langs))
 
-	pool.map(update_translations_p, args)
+	executor.map(update_translations_p, args)
 
 
 def download_translations():
--- a/bench/utils.py
+++ b/bench/utils.py
@@ -825,8 +825,8 @@ def download_translations_p():
 	from concurrent.futures import ThreadPoolExecutor
 
 	executor = ThreadPoolExecutor()
-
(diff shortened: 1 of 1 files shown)


ankush:
Can you add rough benchmarks?

Aradhya-Tripathi:
[
process
](url)
Time taken to run the function using multiprocessing
threading
Time taken to run the function using threads

Aradhya-Tripathi:
As you can see there is a significant boost when using threads as compared to multiple processes.

gavindsouza:
> [ process ](url)
> Time taken to run the function using multiprocessing
> threading
> Time taken to run the function using threads

Mention the code snippet and/or command that gave you this output too @Aradhya-Tripathi

gavindsouza:
On another note, translations are [included in frappe](https://github.com/frappe/frappe/tree/develop/frappe/translations) directly for a while now. This looks like legacy code. 

New translation commands exist under `frappe.commands.translate` module. Using this command would make changes in the apps folder. Maybe we should just get rid of these now. 🤔

Aradhya-Tripathi:
> On another note, translations are [included in frappe](https://github.com/frappe/frappe/tree/develop/frappe/translations) directly for a while now. This looks like legacy code.
> 
> New translation commands exist under `frappe.commands.translate` module. Using this command would make changes in the apps folder. Maybe we should just get rid of these now.

image
there one uses processes and one uses threads
```
def download_translations_t():
	## Refractored code

	from concurrent.futures import ThreadPoolExecutor

	start = time.time()

	with ThreadPoolExecutor() as pool:
		langs = get_langs()
		apps = ('frappe', 'erpnext')
		args = list(itertools.product(apps, langs))

		pool.map(update_translations_p, args)
	end = time.time()

	print(f"TIME TAKEN BY REFRACTORED FUNCTION {end-start}")
```

gavindsouza:
Closing this as the function should be deprecated now instead of optimising it.

PR #1189, opened 2021-07-19
feat: Set default version for easy install to v13

abhishekbalam:
1. Updated default branch while cloning to `version-13`
2. ~~Updated Node version to 14 for v13 requirements~~
3. Updated test case for switch branch command too as it cant switch to a branch which isn't pulled

### DIFF
--- a/bench/playbooks/roles/nodejs/defaults/main.yml
+++ b/bench/playbooks/roles/nodejs/defaults/main.yml
@@ -1,3 +1,3 @@
 ---
-node_version: 14
+node_version: 12
 ...
--- a/bench/playbooks/roles/nodejs/defaults/main.yml
+++ b/bench/playbooks/roles/nodejs/defaults/main.yml
@@ -1,3 +1,3 @@
 ---
-node_version: 12
+node_version: 14
 ...
--- a/bench/tests/test_base.py
+++ b/bench/tests/test_base.py
@@ -15,7 +15,7 @@ import bench.utils
 if sys.version_info.major == 2:
 	FRAPPE_BRANCH = "version-12"
 else:
-	FRAPPE_BRANCH = "version-13"
+	FRAPPE_BRANCH = "develop"
 
 class TestBenchBase(unittest.TestCase):
 	def setUp(self):
--- a/bench/tests/test_base.py
+++ b/bench/tests/test_base.py
@@ -13,9 +13,9 @@ import bench
 import bench.utils
 
 if sys.version_info.major == 2:
-	FRAPPE_BRANCH = "version-13"
+	FRAPPE_BRANCH = "version-12"
 else:
-	FRAPPE_BRANCH = "develop"
+	FRAPPE_BRANCH = "version-13"
 
 class TestBenchBase(unittest.TestCase):
 	def setUp(self):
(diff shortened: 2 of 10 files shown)


bench/tests/test_base.py
```diff
@@ -13,7 +13,7 @@
 import bench.utils
 
 if sys.version_info.major == 2:
-	FRAPPE_BRANCH = "version-12"
+	FRAPPE_BRANCH = "version-13"
```
gavindsouza: Why setup v13 Frappe for PY2? They're incompatible.

bench/playbooks/roles/nodejs/defaults/main.yml
```diff
@@ -1,3 +1,3 @@
 ---
-node_version: 12
+node_version: 14
```
gavindsouza: Does this work with Frappe v12 installs?

bench/tests/test_base.py
abhishekbalam: Right, will make changes

bench/playbooks/roles/nodejs/defaults/main.yml
abhishekbalam: Will test

abhishekbalam: @gavindsouza 

I've left it as 12 as central runs v13 and v12 both with this.

bench/tests/test_base.py
abhishekbalam: Changed develop to v13 instead

bench/tests/test_base.py
```diff
@@ -15,7 +15,7 @@
 if sys.version_info.major == 2:
 	FRAPPE_BRANCH = "version-12"
 else:
-	FRAPPE_BRANCH = "develop"
+	FRAPPE_BRANCH = "version-13"
```
gavindsouza: This removes any way of knowing if bench is compatible with the latest Frappe develop. We might have to add more tests to ensure that.

What is the reason for doing this?

abhishekbalam: The idea was the latest stable version should be tested on priority over develop as that's what would be used most

If we want to write more tests for each version we can, but the tests will need to be updated for newer versions develop aside.

What do you suggest we do? Should I just keep it develop? @gavindsouza

gavindsouza: > The idea was the latest stable version should be tested on priority over develop as that's what would be used most


For testing version compatibility of Frappe with the Bench CLI, I think we still do test the relevant bits (bench creation, requirements resolution, assets building) in the Frappe CI. Anyway, Frappe and ERPNext CIs test this on push and branch merges daily, as compared to the relatively low frequency tests being run here.


> If we want to write more tests for each version we can, but the tests will need to be updated for newer versions develop aside.


Not sure if testing for each major version separately is the way to go. Frappe and Bench are fairly de-coupled in the cases these test suites are written for.


> What do you suggest we do? Should I just keep it develop? @gavindsouza


I'd stick with leaving this as `develop`. Don't have clear enough reasons to make any change.

gavindsouza:
What's the status of this? @abhishekbalam

bench/tests/test_base.py
abhishekbalam: Okay, makes sense

gavindsouza (approved):

PR #1195, opened 2021-08-30
feat: Added option to provide custom queues

fproldan:
What type of a PR is this?

- [x] New Feature Submissions
- [ ] Bug Fix

--------------

This PR adds the option to configure a new settings in `common_site_config.json` that allows creating custom workers besides the defaults hardcoded in `supervisor.conf`.

The new setting is named "workers" and should be in this format:
```json
"workers": {
    "<queue_name>": {
        "timeout": <timeout>, 
        "background_workers": <workers>,
    }   
 }
```

Example:
```json
"workers": {
    "myqueue": {
        "timeout": 5000, 
        "background_workers": 4,
    }   
 }
```

The background_workers param is optional, if not set the general `background_workers` will be used.

### DIFF
--- a/bench/config/supervisor.py
+++ b/bench/config/supervisor.py
@@ -42,7 +42,7 @@ def generate_supervisor_config(bench_path, user=None, yes=False, skip_redis=Fals
 		"background_workers": config.get('background_workers') or 1,
 		"bench_cmd": which('bench'),
 		"skip_redis": skip_redis,
-		"custom_queues": config.get("custom_queues", {}),
+		"workers": config.get("workers", {}),
 	})
 
 	conf_path = os.path.join(bench_path, 'config', 'supervisor.conf')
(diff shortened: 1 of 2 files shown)


gavindsouza (changes requested):
Thanks for your contribution 💯 

I think the config key `workers` would be more accurate instead of `custom_queues`, since they are actually just named workers not queues.

fproldan:
@gavindsouza changes made!

gavindsouza:
LGTM, can you raise a PR that adds docs to `frappe/frappe_docs` too?

gavindsouza (approved):

gavindsouza:
@fproldan We missed adding this to the Procfile 😅 . Can you raise a PR for the same?

fproldan:
@gavindsouza  done!

PR #1196, opened 2021-09-08
feat: `bench setup requirements --dev`

ankush:
Often applications have development or test specific requirements which are not required in production.

Changes:
- Add new flag `bench setup requirements --dev`
- installs all `dev-requirements.txt` in root folder of apps.
- While doing `bench get-app` if dev mode is enabled in common_site_config then dev requirements are also installed. 

Documentation: https://github.com/frappe/frappe_docs/pull/196

### DIFF
--- a/bench/commands/setup.py
+++ b/bench/commands/setup.py
@@ -135,7 +135,7 @@ def setup_socketio():
 @click.command("requirements", help="Setup Python and Node dependencies")
 @click.option("--node", help="Update only Node packages", default=False, is_flag=True)
 @click.option("--python", help="Update only Python packages", default=False, is_flag=True)
-@click.option("--dev", help="Install optional development dependencies", default=False, is_flag=True)
+@click.option("--dev", help="Install optional python development dependencies", default=False, is_flag=True)
 def setup_requirements(node=False, python=False, dev=False):
 	if not (node or python):
 		from bench.utils import update_requirements
@@ -152,6 +152,9 @@ def setup_requirements(node=False, python=False, dev=False):
 	if dev:
 		from bench.utils import install_python_dev_dependencies
 		install_python_dev_dependencies()
+
+		if node:
+			click.secho("--dev flag only supports python dependencies. All node development dependencies are installed by default.", fg="yellow")
 
 
 @click.command("manager", help="Setup bench-manager.local site with the bench_manager app installed on it")
--- a/bench/commands/setup.py
+++ b/bench/commands/setup.py
@@ -135,7 +135,8 @@ def setup_socketio():
 @click.command("requirements", help="Setup Python and Node dependencies")
 @click.option("--node", help="Update only Node packages", default=False, is_flag=True)
 @click.option("--python", help="Update only Python packages", default=False, is_flag=True)
-def setup_requirements(node=False, python=False):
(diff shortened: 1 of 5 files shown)


bench/utils.py
```diff
@@ -617,6 +617,26 @@ def update_node_packages(bench_path='.'):
...
+
+		if os.path.exists(dev_requirements_path):
+			print(f'\n{color.yellow}Installing development dependencies for {app}{color.nc}')
+			exec_cmd(f"{env_py} -m pip install -q -r {dev_requirements_path}", cwd=bench_path)
+		else:
+			print(f'\n{color.red} dev-requirements.txt not found in {app}{color.nc}')
+
+
+def _get_venv_python(bench_path="."):
+	return os.path.join(bench_path, "env", "bin", "python")
```
gavindsouza: There's `bench.utils.get_env_cmd("python")` for this

bench/utils.py
```diff
@@ -617,6 +617,26 @@ def update_node_packages(bench_path='.'):
...
+	env_py = _get_venv_python(bench_path)
+	for app in get_apps():
+		app_path = os.path.join(bench_path, "apps", app)
+		dev_requirements_path = os.path.join(app_path, "dev-requirements.txt")
+
+		if os.path.exists(dev_requirements_path):
+			print(f'\n{color.yellow}Installing development dependencies for {app}{color.nc}')
+			exec_cmd(f"{env_py} -m pip install -q -r {dev_requirements_path}", cwd=bench_path)
+		else:
+			print(f'\n{color.red} dev-requirements.txt not found in {app}{color.nc}')
```
gavindsouza: Use `bench.utils.log` instead. Common interface to log bench commands etc and add colour.

bench/commands/setup.py
```diff
@@ -148,6 +148,11 @@ def setup_requirements(node=False, python=False):
 		from bench.utils import update_node_packages
 		update_node_packages()
 
+@click.command("dev-requirements", help="Setup optional development dependencies")
```
gavindsouza: I think we don't need another command for this. This can be a part of the `setup-requirements` command.

Eg: `bench setup requirements --dev` to setup dev + prod requirements it for all bench apps; fulfilling Python and Node dependencies. `bench setup requirements --dev --python` for only the Python dependencies, etc.

gavindsouza (commented):
This reminded me of back when I tried reducing node_modules size by ignoring the dev requirements in prod. Maybe in the future 🥲

https://github.com/frappe/bench/pull/922

bench/commands/setup.py
ankush: Not sure if we can handle `node`, as dev requirements are installed right now and changing that behaviour could be a breaking change. 😐 

Unlike python req.txt where we are proposing a new separate file, node deps are all stored in `package.json`.

gavindsouza: We could probably though...not for Frappe, but for other apps. Else, we could just put a yellow line/comment saying `--dev` doesn't support fulfilling Node dev deps yet?

ankush: Something along lines of "node dev deps are installed by default"

gavindsouza (approved):

PR #1205, opened 2021-10-18
feat: add `live_reload` to default `common_site_config`

pruthvi145:
## Changes Made
 - add `live_reload` to default_config
 - add patch to set `live_reload` for existing bench instances

---
**Related PR:** https://github.com/frappe/frappe/pull/14238

### DIFF
--- a/patches.txt
+++ b/patches.txt
@@ -1,1 +1,0 @@
-bench.patches.v3.deprecate_old_config
--- a/patches.txt
+++ b/patches.txt
@@ -1,1 +1,2 @@
 bench.patches.v3.deprecate_old_config
+bench.patches.v5.set_live_reload_config
--- a/patches.txt
+++ b/patches.txt
@@ -1,2 +1,1 @@
 bench.patches.v3.deprecate_old_config
-bench.patches.v5.set_live_reload_config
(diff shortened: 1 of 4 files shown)


patches.txt
```diff
@@ -1 +1 @@
-bench.patches.v3.deprecate_old_config
+bench.patches.v3.deprecate_old_config
```
gavindsouza: I think we can get rid of this file altogether. Will leave no scope for confusion

gavindsouza (commented):
This will add the `live_reload` to the common site conf for benches of all Frappe versions. I hope this doesn't break or cause unwanted behaviour for someone else's apps :')

Acknowledging that risk, LGTM.

gavindsouza:
I'm not sure why Travis hasn't finished running. @pruthvi145 Push another commit to re-trigger it? Perhaps drop the `patches.txt` file?

patches.txt
pruthvi145: Done

sagarvora:
> I hope this doesn't break or cause unwanted behaviour for someone else's apps :')

Doesn't `common_site_config` always come with this caveat :smile:
Although, we could do something like this:
```python
frappe_branch = get_current_branch("frappe", bench_path)
if frappe_branch.startswith(("version-11", "version-12", "version-13")):
    return
```

I don't think we should though. If and when they do upgrade to an eligible version, we want the config to be set.

gavindsouza (approved):

PR #1208, opened 2021-10-20
feat: get-app on steroids

gavindsouza:
When you want someone to check out your app, you'd want to reduce as much resistance as possible. Installing and trying out a Frappe App should be a lot easier. One way to make this easier is to get rid of the setting up bench step altogether. Maybe installing the latest version of ERPNext Healthcare on your device can just be:

```bash
bench get healthcare@version-14
```

This command should set up a bench with the latest compatible versions of Frappe and ERPNext respectively. What if you want to use a fork of this? Or, another application not currently under the GitHub frappe org? Consider for instance you want to fetch `mymi14s/language_translator` if the repo https://github.com/mymi14s/language_translator is publicly available. Essentially, you can get any public GitHub repo on your bench by running `bench get` against the `{org}/{repo}@{tag}` pattern.

Each of the following lines means the same exact thing, given that `master` is the default branch for that repo.

```bash
bench get mymi14s/language_translator
bench get mymi14s/language_translator@master
bench get https://github.com/mymi14s/language_translator --branch master
bench get git@github.com:mymi14s/language_translator --branch master
bench get-app git@github.com:mymi14s/language_translator --branch master
```

<details>
<summary><strong>bench get healthcare@develop --init-bench</strong></summary>
<p>

</img>

Note: --init-bench option was added after this image was generated. Check [#issuecomment-981640162](https://github.com/frappe/bench/pull/1208#issuecomment-981640162) for more information.

</p>
</details>

<details><summary><strong>bench --use-feature='dynamic-feed' get healthcare@develop --init-bench</strong></summary>
<p>

Note: --init-bench option was added after this image was generated. Check [#issuecomment-981640162](https://github.com/frappe/bench/pull/1208#issuecomment-981640162) for more information.

A short version that shows how the jobs and individual steps are represented. For more, check this out locally :trollface: :
ezgif com-gif-maker
_(Okay okay...the command said healthcare@develop, but here it says wiki. But I've created ~100 benches in the process of writing this, and this is where I stop LOL)_

Now, this is a somewhat complete get, but yet again for wiki....healthcare takes a really long time. I'd urge you to try installing healthcare locally only ;) 

https://user-images.githubusercontent.com/36654812/142601529-a5af7555-acc9-4976-8118-893467586a56.mov

</p>
</details>

Expanded this to some other commands too. Like, setup requirements for instance:
Screenshot 2021-11-26 at 4 22 56 PM

### Features

- Setup bench if required _(as mentioned above)_
- Pass --use-feature='dynamic-feed' for new CLI.
- Use the process manager of your dreams :P 
  - bench start takes an option now; Use `bench start -m overmind` to use overmind to run your processes
  - `bench restart` will work in developer mode too, if overmind is used
- Setup Frappe app dependencies - using Frappe's `required_apps`
- Added command to drop benches - `bench drop {bench_path}` 
- Aliased commands 
   - `get` for get-app
   - `remove`, `rm` for remove-app
- Utils: `get_available_folder_name`, `get_traceback`
- Renamed `--frappe-branch` option in `bench init` to `--version`. This is because the input may be any git ref: tag, branch. Both options exist though ;)
- Annoyed with all the "Bench is installed in editable..." warnings? Watch them disappear by setting the `BENCH_DEVELOPER` envvar

### Refactors

- Bench's entry point _(cli.py)_ - Reduce de-clutter, consistent behaviour, simplify workflow
- `bench.utils.log` - Maintain `bench.LOG_BUFFER`, deprecate color & use click instead
- Restructured `utils.py` into a module with separate files: app, bench, cli, render, system, translation
- `switch-branch` - Workflow optimization, better readability
- `update` - Separate validations, make function de-coupled with the CLI, use newer APIs
- Setting up dev requirements - Remove inconsistent logs, fix order of resolution
- `get_app` - Fix path, cwd issues and order of resolving deps: py -> dev py -> yarn, fix verbosity
- All the code that the new `Bench`, `App` object notations can take care of  ;)

### Bug Fixes

- Show pip's help menu when --help is passed in pip command
- Show traceback if an error occurs during init, get outside bench dir
- raise `CommandFailedError` (by default) when `exec_cmd` fails
- Show tracebacks on bench command failures, unconditionally

### Other Changes

- Added a CLI Rendering Engine - f230778, f117959
- Object Notations for Bench, App & more - check _bench.py_, _app.py_
- Rename conflicting folders during app archiving, bench inits
- Aliased commands - bench now inherits `MultiCommandGroup` that allows adding multiple aliases for the same command.
- Add compatibility and license shields to README
- Styled code based on own preference, flake8 & Black
- Dropped dead code - `bench.prepare_staging`
- Dropped support for Frappe v4-5-6 - Use `pip install frappe-bench --only-binary='all'` to install the latest supported version for your machine :D
- Deprecated `bench setup socketio` - Frappe has been managing this for a while now
- CI: Use Nodejs version 14 for Tests


vjFaLk:
@gavindsouza  I like the idea of being able to quickly provision a bench around an app (or a branch of that app), however, I don't think it should replace get-app, but instead be a separate command altogether (and not even have `get` in it to avoid confusion). A command behaving completely differently depending on where it's being run, I think is confusing. The behavior is too implicit IMO.

gavindsouza:
> @gavindsouza I like the idea of being able to quickly provision a bench around an app (or a branch of that app), however, I don't think it should replace get-app, but instead be a separate command altogether (and not even have `get` in it to avoid confusion).

The goal is to make installing Frappe Apps with more ease, especially for devs new to this ecosystem. This PR makes two changes in get-app: 
1. Initializes bench if it doesn't already exist.
2. Installs dependent Frappe Apps prior to app install. The goal is to broaden this and manage resolving app versions too. 

The bench concept isn't very obvious for most Frappe newbies, and these changes aim to get rid of that extra confusion and reduce the steep learning curve. 

> A command behaving completely differently depending on where it's being run, I think is confusing. The behavior is too implicit IMO.

I'm not sure if this is a problem. Doesn't seem to be to me since I think the proposed way optimizes for DX. But if there are more voices on this matter, we could reach some consensus. 

**Edit:** Maybe an extra flag like `--init-bench` to the `get-app` command would work instead? Without it, command fails saying not in bench, but also run with `--init-bench` if you want to create one.

---

If we choose against what we already have here, do suggest a command name for the same. _(If `get` and `get-app` are to be two different commands, one with the newer stuff and the other with the same old behavior (just clone+install), that'd be too confusing)_

gavindsouza:
I've gotten rid of the implicit bench initialization. Now, you can pass `--init-bench` in bench get(-app) to initialize bench, else a `NotInBenchDirectoryError` exception is raised. 
Screenshot 2021-11-29 at 7 02 25 PM

Also, added a `--install-app` option in the bench init command. So, command `get-app` + option `--init-bench` is the same as command `init` + option `--install-app`

vjFaLk:
@gavindsouza Sorry for the late reply, this got buried in my emails. Appreciate the more explicit flag!

PR #1218, opened 2021-11-19
feat: add no-git option to new-app command

### LINKED ISSUE #1173: bench new-app --no-git
Issue: Feature Request

bench new-app creates a new git repo

`bench new-app --no-git app_name` should create app with no git repo.

**Describe alternatives you've considered**

```shell
bench new-app app_name
rm -rf apps/app_name/.git
```

phot0n:
Add `--no-git` option in `new-app` command for initialising non-git apps

closes: https://github.com/frappe/bench/issues/1173
docs: https://github.com/frappe/frappe_docs/pull/226

### DIFF
--- a/bench/commands/make.py
+++ b/bench/commands/make.py
@@ -149,11 +149,17 @@ def get_app(
 
 
 @click.command("new-app", help="Create a new Frappe application under apps folder")
+@click.option(
+	"--no-git",
+	is_flag=True,
+	flag_value="--no-git",
+	help="Do not initialize git repository for the app (available in Frappe v14+)"
+)
 @click.argument("app-name")
-def new_app(app_name):
+def new_app(app_name, no_git=None):
 	from bench.app import new_app
 
-	new_app(app_name)
+	new_app(app_name, no_git)
 
 
 @click.command(
--- a/bench/app.py
+++ b/bench/app.py
@@ -355,12 +355,27 @@ def get_app(
 		app.install(verbose=verbose, skip_assets=skip_assets)
 
 
-def new_app(app, bench_path="."):
+def new_app(app, no_git=None, bench_path="."):
+	if bench.FRAPPE_VERSION in (0, None):
+		raise NotInBenchDirectoryError(
+			f"{os.path.realpath(bench_path)} is not a valid bench directory."
+		)
+
 	# For backwards compatibility
 	app = app.lower().replace(" ", "_").replace("-", "_")
-	logger.log(f"creating new app {app}")
 	apps = os.path.abspath(os.path.join(bench_path, "apps"))
-	run_frappe_cmd("make-app", apps, app, bench_path=bench_path)
+	args = ["make-app", apps, app]
+	if no_git:
+		if bench.FRAPPE_VERSION < 14:
+			click.secho(
+				"Frappe v14 or greater is needed for '--no-git' flag",
+				fg="red"
+			)
+			return
+		args.append(no_git)
+
+	logger.log(f"creating new app {app}")
+	run_frappe_cmd(*args, bench_path=bench_path)
 	install_app(app, bench_path=bench_path)
 
 


bench/commands/make.py
```diff
@@ -64,9 +64,10 @@ def get_app(git_url, branch, name=None, overwrite=False, skip_assets=False):
 
 @click.command('new-app', help='Create a new Frappe application under apps folder')
 @click.argument('app-name')
-def new_app(app_name):
+@click.option('--no-git', is_flag=True, flag_value="--no-git", help='Initialize the app not to be a git repo')
```
surajshetty3416:
```suggestion
@click.option('--no-git', is_flag=True, flag_value="--no-git", help='Do not initialize git repository for the app')
```

surajshetty3416:
@gavindsouza this will only work with develop... do have a system to check the compatible versions?

gavindsouza:
I think we'll need to do a couple of things:
* Mention in the help that `--no-git` flag is v14+ only
* Check Frappe version using `get_current_version('frappe')`

phot0n:
it looks like this now:

Screenshot 2021-11-24 at 1 54 05 PM

gavindsouza:
@phot0n Can you resolve the conflicts?

phot0n:
also, is my review visible @gavindsouza ? - i had added a review not sure why it's not visible (tried opening it on incognito) but here it goes:

had to add [this](https://github.com/frappe/bench/pull/1218/files#diff-3722f4ad716221f5c6d162bf82efd37d96c6367678903aba4e51725dd57ee75bR169) check because of this:

Screenshot 2021-11-24 at 3 17 21 PM

even though we're not in bench dir, it's still printing out the v14 message.
ideally (i think) we should not continue with execution if it's not bench dir ..can i know the reason why we're printing the WARN-ing on console but not killing the process after that? [ref](https://github.com/frappe/bench/blob/b9b1d444b1d340ce43fe300b7350cb8d76fcc283/bench/cli.py#L70)

gavindsouza:
> also, is my review visible @gavindsouza ? - i had added a review not sure why it's not visible (tried opening it on incognito)

Likely because you haven't submitted the review? It must be in Pending state.


> ideally (i think) we should not continue with execution if it's not bench dir ..can i know the reason why we're printing the WARN-ing on console but not killing the process after that? [ref](https://github.com/frappe/bench/blob/b9b1d444b1d340ce43fe300b7350cb8d76fcc283/bench/cli.py#L70)

The initial problem was that people ran commands in the "wrong" working directories, bench didn't provide any guidance and they'd be lost wondering what's wrong. There are two groups of commands that can be run outside bench directories...not sure if there's any intersection in the other groups. 

The warning was added as a nudge. There's no reason to not show it as an error instead and kill the process, but doing it correctly is a separate task. While being mindful of whether changing the behaviour of a particular command could break someone else's workflow, the easier way to nudge people was by just throwing a warning message that they can ignore and go ahead with their lives.


> had to add [this](https://github.com/frappe/bench/pull/1218/files#diff-3722f4ad716221f5c6d162bf82efd37d96c6367678903aba4e51725dd57ee75bR169) check because of this:
> 
> Screenshot 2021-11-24 at 3 17 21 PM

I'd suggest you raise `NotInBenchDirectoryError` instead of returning though, which was recently added via #1208

https://github.com/frappe/bench/blob/357fcc0bbd7e0ea6f7546d61c5223f1288b89517/bench/exceptions.py#L31

gavindsouza:
Another thing, you don't need to run `get_current_frappe_version` actually, you can find out the Frappe version by accessing `bench.FRAPPE_VERSION` directly :)

If it's value is 0 or None, you're outside of a bench dir, else you're in one...no need for any other checks either.

phot0n:
@gavindsouza I've rebased it to the current develop branch and made the changes as well. can you have a look?

gavindsouza (approved):

PR #1221, opened 2021-12-01
chore: change the default version to 13 of erpnext and frappe when installing using easyinstall

phot0n:
In a meet with @abhishekbalam - we got confused and figured out why easyinstall was installing frappe/erpnext v12 by default

This pr bumps the default version of frappe & erpnext to 13 when installing via easyinstall script

PS: I haven't tested this as i don't have ansible installed on my local machine
PPS: we should come up with some other way I think - as more versions get released we'll have to do this everytime to keep up with the releases [or maybe we can mention it in readme and/or [docs](https://frappeframework.com/docs/v13/user/en/production-setup) about the `--version` argument?]

### DIFF
--- a/install.py
+++ b/install.py
@@ -251,16 +251,13 @@ def install_bench(args):
 	if args.production:
 		extra_vars.update(max_worker_connections=multiprocessing.cpu_count() * 1024)
 
-	frappe_branch = 'version-13'
-	erpnext_branch = 'version-13'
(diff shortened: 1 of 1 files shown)


phot0n:
we could make the `--version` flag a mandatory thing? - that way we won't have to worry about bumping the default version (?)

gavindsouza (approved):

PR #1227, opened 2021-12-07
fix: process SSH git URLs

Alchez:
Trying to run the following command: `bench get-app git@github.com:org/repo.git` would result in the URL being transformed to `https://github.com/git@github.com:org/repo.git`, which errors out and `get-app` fails.

**Alternatives:**
This isn't the best way to handle this, and we could use something like [giturlparse](https://github.com/nephila/giturlparse). But I didn't want to add a dependency right now without getting some feedback.

### DIFF
--- a/bench/app.py
+++ b/bench/app.py
@@ -84,6 +84,8 @@ class AppMeta:
 
 		# fetch meta for repo from remote git server - traditional get-app url
 		elif is_git_url(self.name):
+			if self.name.startswith("git@") or self.name.startswith("ssh://"):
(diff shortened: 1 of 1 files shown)


ankush:
IIRC this used to work before, I almost exclusively use ssh URLs for all the apps. Not sure what broke it.

Alchez:
@ankush, I think the rework branched out the logic for processing git URLs, which is good design, but the SSH side wasn't setup at this place

ankush (approved):
Works as expected.

PR #1233, opened 2021-12-08
fix: get-app command should be able to clone from any git server

### LINKED ISSUE #1228: Bench get-app always use github.com even if we use different git server such as gitlab
Issue: Bug report

To reproduce, run bench get-app (anything from gitlab) 

An error, will appear: 
fatal: unable to access 'https://github.com/git@gitlab.com:....

Looking at the codes it seems that the remote_server is set to github.com but I can't find where it's actually set to a custom url.
Therefore when running bench get-app from anything other than gitlab it just don't work anymore.

Apology of course if I'm not using the command right, however this was working a while ago, last known working version was 5.0.0-dev.

Thanks a lot for your help.

phot0n:
Currently the git remote server is hardcoded to `github.com`.

Changes:
- get-app command now has 2 new options `--git-host` (for providing from which git host to get the app from - currently only supported for gitlab and github[default]) and `--use-ssh` - both of these options are to be used with partial url format (previously name tag format) i.e `org/repo@branch` or `org/repo` or `repo@branch` or just `repo` (the last 2 being for apps under frappe/erpnext org) 
- support for `org/repo` format in get-app (rest all partial url formats were already there)
- fix for get-app not being able to get app from anywhere but github - if the url provided is valid, we just use it as is for cloning
- added a new dependency [`git-url-parse`](https://github.com/coala/git-url-parse) - which we also use in frappe

closes: https://github.com/frappe/bench/issues/1228

Screenshot 2021-12-08 at 8 01 26 PM

Screenshot 2021-12-10 at 2 49 55 PM

Screenshot 2021-12-10 at 2 49 03 PM

Screenshot 2021-12-10 at 4 43 37 PM

Screenshot 2021-12-10 at 5 02 02 PM

### DIFF
--- a/bench/app.py
+++ b/bench/app.py
@@ -43,7 +43,6 @@ class AppMeta:
 		self,
 		name: str,
 		branch: str = None,
-		git_host: str = None,
 		to_clone: bool = True,
 		use_ssh: bool = False
 	):
@@ -51,10 +50,11 @@ class AppMeta:
 		name (str): This could look something like
 			1. https://github.com/frappe/healthcare.git
 			2. git@github.com:frappe/healthcare.git
-			3. frappe/healthcare@develop
-			4. frappe/healthcare
-			5. healthcare
+			3. github:frappe/healthcare@develop
+			4. frappe/healthcare@develop
+			5. frappe/healthcare
 			6. healthcare@develop, healthcare@v13.12.1
+			7. healthcare
 
 		References for Version Identifiers:
 		 * https://www.python.org/dev/peps/pep-0440/#version-specifiers
@@ -70,7 +70,6 @@ class AppMeta:
 
 		# for partial url format
 		self.is_partial_url = False
-		self.git_host = git_host
 		self.use_ssh = use_ssh
 
 		self.branch = branch
@@ -94,8 +93,12 @@ class AppMeta:
 			self._setup_details_from_mounted_disk()
 
 		# fetch meta for repo from git url - traditional get-app url
-		# NOTE: Adding a check for owner here for making `organisation/repoository` format work
-		elif parsed_git_obj and parsed_git_obj.owner:
+		# NOTE: additional checks for making `org/repo` and `githost:org/repo`
+		# partial url formats work - basically checking if the protocol is ssh,
+		# there should be a user associated with it, for it to be a valid ssh git url
+		elif parsed_git_obj and not (
+			parsed_git_obj.protocol == "ssh" and not parsed_git_obj.user
+		):
 			self.tag = self.branch
 			self.org = parsed_git_obj.owner
 			self.repo = parsed_git_obj.name
@@ -108,7 +111,7 @@ class AppMeta:
 		self.org, self.repo, self.tag = os.path.split(self.name)[-2:] + (self.branch,)
 
 	def _setup_details_from_partial_url(self):
-		self.org, self.repo, self.tag = fetch_details_from_partial_url(self.name)
+		self.git_host, self.org, self.repo, self.tag = fetch_details_from_partial_url(self.name)
 
 		if self.tag and self.branch:
 			click.secho(
@@ -123,20 +126,6 @@ class AppMeta:
 		self.org, self.repo, self.tag = os.path.split(
 			os.path.join(self.bench.name, "apps", self.name)
 		)[-2:] + (self.branch,)
-
-	def _setup_details_from_git_url(self):
-		return self.__setup_details_from_git()
-
-	def __setup_details_from_git(self):
-		if self.use_ssh:
-			_first_part, _second_part = self.name.split(":")
-			self.remote_server = _first_part.split("@")[-1]
-			self.org, _repo = _second_part.split("/")
-		else:
-			self.remote_server, self.org, _repo = self.name.split("/")[-3:]
-
-		self.tag = self.branch
-		self.repo = _repo.split(".")[0]
 
 	@property
 	def url(self):
@@ -310,7 +299,6 @@ def get_app(
 	git_url,
 	branch=None,
 	bench_path=".",
-	git_host="github",
 	skip_assets=False,
 	verbose=False,
 	overwrite=False,
@@ -329,9 +317,7 @@ def get_app(
 	import bench.cli as bench_cli
 
 	bench = Bench(bench_path)
-	app = App(
-		git_url, branch=branch, bench=bench, git_host=git_host, use_ssh=use_ssh
-	)
+	app = App(git_url, branch=branch, bench=bench, use_ssh=use_ssh)
 	git_url = app.url
 	repo_name = app.repo
 	branch = app.tag
--- a/bench/app.py
+++ b/bench/app.py
@@ -19,10 +19,10 @@ from bench.exceptions import NotInBenchDirectoryError
 from bench.utils import (
 	fetch_details_from_partial_url,
 	get_available_folder_name,
+	get_parsed_giturl,
 	is_bench_directory,
 	log,
-	parse_git_url,
-	run_frappe_cmd,
+	run_frappe_cmd
 )
 from bench.utils.bench import (
 	build_assets,
@@ -76,7 +76,7 @@ class AppMeta:
 		self.setup_details()
 
 	def setup_details(self):
-		parsed_git_obj = parse_git_url(self.name)
+		parsed_git_obj = get_parsed_giturl(self.name)
 
 		# fetch meta from installed apps
 		if (
--- a/bench/app.py
+++ b/bench/app.py
@@ -45,8 +45,9 @@ class AppMeta:
 			1. https://github.com/frappe/healthcare.git
 			2. git@github.com:frappe/healthcare.git
 			3. frappe/healthcare@develop
-			4. healthcare
-			5. healthcare@develop, healthcare@v13.12.1
+			4. frappe/healthcare
+			5. healthcare
+			6. healthcare@develop, healthcare@v13.12.1
 
 		References for Version Identifiers:
 		 * https://www.python.org/dev/peps/pep-0440/#version-specifiers
@@ -67,16 +68,10 @@ class AppMeta:
 		self.setup_details()
 
 	def setup_details(self):
-
-		# fetch meta for repo from git url - traditional get-app url
 		parsed_git_obj = parse_git_url(self.name)
-		if parsed_git_obj:
-			self.tag = self.branch
-			self.org = parsed_git_obj.owner
-			self.repo = parsed_git_obj.name
 
 		# fetch meta from installed apps
-		elif (
+		if (
 			not self.to_clone
 			and hasattr(self, "bench")
 			and os.path.exists(os.path.join(self.bench.name, "apps", self.name))
@@ -88,6 +83,13 @@ class AppMeta:
 		elif os.path.exists(self.name):
 			self.on_disk = True
 			self._setup_details_from_mounted_disk()
+
+		# fetch meta for repo from git url - traditional get-app url
+		# NOTE: Adding a check for owner here for making `organisation/repoository` format work
+		elif parsed_git_obj and parsed_git_obj.owner:
+			self.tag = self.branch
+			self.org = parsed_git_obj.owner
+			self.repo = parsed_git_obj.name
 
 		# fetch meta from new styled name tags & first party apps on github
 		else:
--- a/bench/app.py
+++ b/bench/app.py
@@ -17,7 +17,7 @@ import click
 import bench
 from bench.exceptions import NotInBenchDirectoryError
 from bench.utils import (
-	fetch_details_from_tag,
+	fetch_details_from_partial_url,
 	get_available_folder_name,
 	is_bench_directory,
 	log,
@@ -39,7 +39,14 @@ if typing.TYPE_CHECKING:
 
 
 class AppMeta:
-	def __init__(self, name: str, branch: str = None, to_clone: bool = True):
+	def __init__(
+		self,
+		name: str,
+		branch: str = None,
+		git_host: str = None,
+		to_clone: bool = True,
+		use_ssh: bool = False
+	):
 		"""
 		name (str): This could look something like
 			1. https://github.com/frappe/healthcare.git
@@ -58,12 +65,14 @@ class AppMeta:
 		"""
 		self.name = name.rstrip('/')
 		self.to_clone = to_clone
-		# NOTE: keeping this here in lieu of:
-		# https://github.com/frappe/bench/pull/1233#discussion_r766489037
-		self.remote_git_host = "github.com"
 		self.on_disk = False
-		self.from_name_tag = False
 		self.from_apps = False
+
+		# for partial url format
+		self.is_partial_url = False
+		self.git_host = git_host
+		self.use_ssh = use_ssh
+
 		self.branch = branch
 		self.setup_details()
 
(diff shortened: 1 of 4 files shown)


phot0n:
since @Alchez's [pr](https://github.com/frappe/bench/pull/1227) considers the git url format which starts with `ssh://`, should we add support for that as well?

PS: this format works fine as far as I can test
`ssh://git@hostserver/org/repo.git`
are there other formats starting with `ssh://` for git url?

Alchez:
I would actually suggest using an external dependency ([giturlparse](https://pypi.org/project/giturlparse/)) that has already solved the problem of managing git URLs. That handles everything we need, and even spits out URLs for whatever protocol we need.

phot0n:
@Alchez - so had a chat with @ankush , we already use a git parsing library in frappe - https://github.com/coala/git-url-parse - so it'll be better to use the same :)

Alchez:
@phot0n, yeah that works too.

At this point though, I feel like we need to zoom out a bit and look at the design itself. Because we shouldn't even need to extract the host and just simply pass it to the `git clone` command 😅 

The only place I can see where we could need a URL parser is if we wanted to validate if a provided URL is actually a valid git link before processing it. That's what the Frappe app does for the Github connector atleast.

phot0n:
Agreed, had the same thought as to why we're deconstructing and then reconstructing the git url 😅  - though i think it makes sense on shorter slugs where we just pass the repo name and/or branch/tag instead of fully passing the git url

phot0n:
the https://github.com/frappe/bench/pull/1233/commits/c872181b3009111906b52cb786239608b38cce6c will probably fail on python v3.7 test (might fail on all of them 🙈 ) as i've used walrus operator [over here](https://github.com/frappe/bench/pull/1233/files#diff-3722f4ad716221f5c6d162bf82efd37d96c6367678903aba4e51725dd57ee75bR85) (which was added in python v3.8 - will change it)

I've used the git url parsing dependency we use in frappe and did not change the design on how the App Meta class was implemented.
Please do let me know your thoughts @ankush @Alchez :)

Also, please do let me know if any better name pops up for [this](https://github.com/frappe/bench/pull/1233/files#diff-e8c1559baffd734600ddd1974fd40b907e7e5297a88378f0b808c4f078008e49R423) function :P

bench/utils/__init__.py
```diff
@@ -420,10 +421,14 @@ def fetch_details_from_tag(_tag: str) -> Tuple[str, str, str]:
 	return org, repo, tag
 
 
-def is_git_url(url: str) -> bool:
-	# modified to allow without the tailing .git from https://github.com/jonschlinkert/is-git-url.git
-	pattern = r"(?:git|ssh|https?|\w*@[-\w.]+):(\/\/)?(.*?)(\.git)?(\/?|\#[-\d\w._]+?)$"
-	return bool(re.match(pattern, url))
+def check_or_get_parsed_git_obj_from_url(url: str) -> Any:
```
Alchez: I think this could just be called `parse_url` or `parse_git_url`

Alchez: `Any` is too broad. I think you can either get rid of the return type entirely, or use something like `Union[giturlparse.parse.Parsed, False]`

bench/app.py
```diff
@@ -123,16 +116,14 @@ def url(self):
...
-		return self.get_http_url()
+		# return the git url as is
+		return self.name
 
 	def get_http_url(self):
-		return f"https://{self.remote_server}/{self.org}/{self.repo}.git"
-
-	def get_ssh_url(self):
-		return f"git@{self.remote_server}:{self.org}/{self.repo}.git"
+		return f"https://github.com/{self.org}/{self.repo}.git"
```
Alchez: Doesn't this also assume `github.com` as the host? If you store the parsed URL in the `App` instance, that'll let you access the host name too.

Alchez (changes requested):

bench/app.py
phot0n: So the `get_http_url` method will only be called in the case when `self.from_name_tag` is `True` i.e the cases where user provides ONLY the repo name and/or tag/branch as in these cases we check if the repos belong under frappe/erpnext org (eg: `bench get-app healthcare` or `bench get-app healthcare@develop`) - in the other case where the user provides a (valid) git url, it will just be returned as is as there is no need to do anything else as you've mentioned in one of your comments :)

Alchez: Oh, I see now. In that case, this is fine (you might want to keep and rename the `remote_server` variable and just use that though).

We could also later consider receiving an optional parameter for the host in the CLI (defaults to github):

`bench get-app --host gitlab org/repo`

phot0n:
I've updated the pr description and added some screenshots as well :)

TurkerTunali:
IMHO get-app should use exact git clone address. Because get-app is a developer method. Bench should never use healthcare@develop shortcuts. The developer should be freely use his repo and branch like git clone method. It makes harder to get in frappeverse.

phot0n:
@TurkerTunali that option is not new i think as i saw it in previous bench version as well and is only available for first party apps i.e apps present under frappe/erpnext github org(s) and if a git url is passed it will be treated like how `git clone` treats it (tbh that shortcut is also treated in the same way except we just create the url through code) - as essentially we do use `git clone` for cloning the repos when we use `bench get-app` 😅

TurkerTunali:
@phot0n I think we have a litle bit trouble here, but I assume that development is already done. So we may leave it as it is.

phot0n:
@TurkerTunali please do let me know, i don't mind writing it over again :) - this needs to be correct otherwise people won't be able to get their apps 😆

bench/app.py
phot0n: i read the logic for [`fetch_details_from_tag`](https://github.com/frappe/bench/blob/develop/bench/utils/__init__.py#L403) method and i think it can do it for all the repos on github (atleast for now) - if the format is `bench get-app org/repo@branch` ..i'm shocked 😅 

also, if i do something like `bench get-app org/repo` the git parser picks it up to be a valid url 🙄 - which errors out 😓

Screenshot 2021-12-13 at 7 47 03 PM

I should probably add a check for owner which will allow `bench get-app org/repo` to work correctly as it will go to `fetch_details_from_tag` method

phot0n:
I've added some changes last night, mostly the feature @Alchez  talked about [here](https://github.com/frappe/bench/pull/1233#discussion_r766489037) (got a bit carried away 😅 ) - now we can use `bench get-app org/repo` (and for a specific branch `org/repo@branch` - which was already there) throughout github by default and also gitlab by providing `--git-host gitlab` flag. we can also specify if we want to use ssh while cloning the app with this name tag format by providing `--use-ssh` flag.

Please do let me know your thoughts and also if you want them removed as i guess i went a bit with flow 😅

bench/app.py
```diff
@@ -59,15 +67,18 @@ class Healthcare(AppConfig):
 			dependencies = [{"frappe/erpnext": "~13.17.0"}]
 		"""
 		self.name = name.rstrip('/')
-		self.remote_server = "github.com"
 		self.to_clone = to_clone
 		self.on_disk = False
-		self.use_ssh = False
+		self.from_name_tag = False
+		self.remote_git_host_for_name_tag = host
+		self.use_ssh_for_name_tag = use_ssh
```
Alchez: I think we can improve the variable naming here. Maybe something like:

```suggestion
		self.is_partial_url = False
		self.git_host = host
		self.use_ssh = use_ssh
```

Alchez (changes requested):
@phot0n, I've suggested some changes (plus documentation changes) that are really nitpicky, but otherwise this is looking good.

As a totally optional side, looking at the code now, I like [giturlparse](https://pypi.org/project/giturlparse/) more for us, cause it'll offload the URL validation and generation to the library. But this PR can go as is too.

bench/app.py
```diff
@@ -98,23 +112,21 @@ def _setup_details_from_mounted_disk(self):
 	def _setup_details_from_name_tag(self):
 		self.org, self.repo, self.tag = fetch_details_from_tag(self.name)
 
+		if self.tag and self.branch:
+			click.secho(
+				"Ignoring the branch provided with `--branch` option",
+				fg="bright_yellow"
+			)
+		elif not self.tag and self.branch:
+			self.tag = self.branch
```
phot0n: @Alchez i've added this logic for picking up the branch when partial url is provided - so as we can also provide branch with `--branch` flag, i'm ignoring the that value when the partial url has branch provided in itself (i.e `org/repo@branch`) - otherwise when the partial url doesn't have the branch provided but is provided by `--branch` flag, we take the flag's value

this can be shortened to `self.tag = self.tag or self.branch` but i wanted to show that message so the the user knows what's happening 😅

Alchez: @phot0n, yeah that makes sense. I would also suggest changing any function names that have `name-tag` to `partial-url`, to keep it consistent

phot0n: done :)
Please do have a look ...

Alchez: Yeah, looks good now

bench/app.py
```diff
@@ -79,39 +93,37 @@ def setup_details(self):
...
-		elif is_git_url(self.name):
-			if self.name.startswith("git@") or self.name.startswith("ssh://"):
-				self.use_ssh = True
-			self._setup_details_from_git_url()
+		# fetch meta for repo from git url - traditional get-app url
+		# NOTE: Adding a check for owner here for making `organisation/repoository` format work
+		elif parsed_git_obj and parsed_git_obj.owner:
+			self.tag = self.branch
+			self.org = parsed_git_obj.owner
+			self.repo = parsed_git_obj.name
```
gavindsouza, line 104: We could also set `self.use_ssh` if the passed name was a git SSH url?

bench/app.py
```diff
@@ -39,14 +39,22 @@
...
 		"""
 		name (str): This could look something like
 			1. https://github.com/frappe/healthcare.git
 			2. git@github.com:frappe/healthcare.git
 			3. frappe/healthcare@develop
-			4. healthcare
-			5. healthcare@develop, healthcare@v13.12.1
+			4. frappe/healthcare
+			5. healthcare
+			6. healthcare@develop, healthcare@v13.12.1
```
gavindsouza, line 56: I think we can add support for `github:frappe/healthcare@version-13` too? SInce we want the gitserver host to be available too 🤔 . We could allow passing TLDs too instead of hardcoding them unconditionally in the `get_*_url` methods. Eg: `git.gav.in:frappe/healthcare@version-13` would also be valid.

bench/utils/__init__.py
```diff
@@ -420,10 +421,14 @@ def fetch_details_from_tag(_tag: str) -> Tuple[str, str, str]:
...
-def is_git_url(url: str) -> bool:
-	# modified to allow without the tailing .git from https://github.com/jonschlinkert/is-git-url.git
-	pattern = r"(?:git|ssh|https?|\w*@[-\w.]+):(\/\/)?(.*?)(\.git)?(\/?|\#[-\d\w._]+?)$"
-	return bool(re.match(pattern, url))
+def parse_git_url(url: str):
+	try:
+		parsed_obj = giturlparse.parse(url)
+	except Exception:
+		# the git url is not parsable
+		return False
```
gavindsouza: Why not let it just raise instead? That's what I'd expect from a util named `parse_git_url`.

gavindsouza (changes requested):

bench/app.py
phot0n: it's not necessary i think - it's there to go with the partial url format (when user provides `--use-ssh` - maybe we can change the naming a bit as it definitely is a bit confusing as to with what it might be used [maybe `use_ssh_for_partial_url`?])

bench/utils/__init__.py
phot0n: if we look at how it's being [used](https://github.com/frappe/bench/pull/1233/files#diff-3722f4ad716221f5c6d162bf82efd37d96c6367678903aba4e51725dd57ee75bR80) - we'll see that we pass everything coming our way into `parse_git_url` ..so if the parser raises everytime we'll need to handle that. we can shift the exception handling to `setup_details` method (?)

bench/app.py
phot0n: tbh i'm not sure ..i was thinking of removing the one i've added as it seemed a bit excessive to me (kinda went with the flow 😅 ) - though if you want it i can try to add that but the host one is kinda solved using `--git-host` option which is also only for using with partial urls

gavindsouza: Passing `--git-host` is one solution that seems to miss out on the whole idea of why this format was added. 

We should get rid of the git_host kwarg altogether. Just fetch this by parsing the url. That's how partial URLs should be handled. Adding a new option doesn't work while integrating this system in the App Dependency System.

bench/utils/__init__.py
gavindsouza: The problem I was trying to address here was a naming one. Everything could stay wherever they are if this function is renamed to `get_parsed_giturl` and returns `None` instead of `False`.

gavindsouza (changes requested):

bench/utils/__init__.py
phot0n: ah sorry i misinterpreted 😅 - will do 👍

bench/app.py
phot0n: hmm, so basically the we want to get the host from the partial format itself rather than providing it through an option - we can do that though i'm not sure how much different they are though cause they basically achieve the same thing 🤔

Alchez: > I think we can add support for `github:frappe/healthcare@version-13` too? SInce we want the gitserver host to be available too 🤔 . We could allow passing TLDs too instead of hardcoding them unconditionally in the `get_*_url` methods. Eg: `git.gav.in:frappe/healthcare@version-13` would also be valid.

If we end up supporting hosts in that partial fragment too, I think at that point we might as well expect people to use the full URL anyway, no?

The new `get-app` is a good option to use when some of the things expected in the URL can be defaulted (host to github, ref to default branch/tag). IMHO, anything beyond that should be received either in the full URL, or as separate options.

phot0n: +1 for alchez's comment - i think most people don't really care about shortcuts like these - we/they just go to a repo and get the git url and start doing things

gavindsouza:
I believe the issue this PR was addressing is fixed via https://github.com/frappe/bench/pull/1243. Do we need to salvage anything from here?

phot0n:
@gavindsouza if we have a look at https://github.com/frappe/frappe/issues/15229 - the git url being used there is `https://dev.azure.com/user/custom/_git/custom_crm.git` which will again fail with the changes in #1243 due to the reason that we're decomposing/recomposing urls (based on the assumption that every url will have the same structure) - it's better to just return/use the url as is if it's a valid git url

Screenshot 2021-12-30 at 4 53 42 PM
(tried on latest develop)

gavindsouza:
Seems like giturlparse can't handle azure links either (ref: https://github.com/nephila/giturlparse/issues/6, testing both libraries locally). Raised https://github.com/frappe/bench/pull/1247 to fix the same.

Edit: Updated after realizing Oopsie

Alchez:
> Seems like giturlparse can't handle azure links either (ref: [nephila/giturlparse#6](https://github.com/nephila/giturlparse/issues/6)). Raised #1247 to fix the same.

Just a note cause it seems like package names are to blame here, You've used code from the https://github.com/coala/git-url-parse (v1.2.2) library over at https://github.com/nephila/giturlparse/issues/6 (v0.10.0).

Both of these libraries seem to have their own issues though.

phot0n:
closing this pr in lieu of #1247 - thanks for all the help Alchez, Gavin :)

PR #1235, opened 2021-12-09
fix: PosixPath object has no attribute 'rstrip'

### LINKED ISSUE #1234: AttributeError: 'PosixPath' object has no attribute 'rstrip'
Bench update failing due to a patch.
```
ERROR: 'PosixPath' object has no attribute 'rstrip'
  File "/usr/local/bin/bench", line 33, in <module>
    sys.exit(load_entry_point('frappe-bench', 'console_scripts', 'bench')())
  File "/home/frappe/bench/bench/cli.py", line 117, in cli
    raise e
  File "/home/frappe/bench/bench/cli.py", line 107, in cli
    bench_command()
  File "/usr/local/lib/python3.6/dist-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.6/dist-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.6/dist-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.6/dist-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.6/dist-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/home/frappe/bench/bench/commands/update.py", line 23, in update
    update(pull=pull, apps=apps, patch=patch, build=build, requirements=requirements, restart_supervisor=restart_supervisor, restart_systemd=restart_systemd, backup=not no_backup, compile=not no_compile, force=force, reset=reset)
  File "/home/frappe/bench/bench/utils/bench.py", line 363, in update
    patches.run(bench_path=bench_path)
  File "/home/frappe/bench/bench/patches/__init__.py", line 21, in run
    result = execute(bench_path)
  File "/home/frappe/bench/bench/patches/v5/update_archived_sites.py", line 43, in execute
    shutil.move(archived_site_path, new_directory)
  File "/usr/lib/python3.6/shutil.py", line 546, in move
    real_dst = os.path.join(dst, _basename(src))
  File "/usr/lib/python3.6/shutil.py", line 510, in _basename
    return os.path.basename(path.rstrip(sep))
AttributeError: 'PosixPath' object has no attribute 'rstrip'
```

Line 43 of update_archived_sites, `shutil.move(archived_site_path, new_directory)` throws this error.

rtdany10:
~Convert paths to strings before passing them to `shutil.move()`.~
Drop `shutil` and use `Path.rename` instead.
Closes #1234

### DIFF
--- a/bench/patches/v5/update_archived_sites.py
+++ b/bench/patches/v5/update_archived_sites.py
@@ -40,7 +40,7 @@ def execute(bench_path):
 		os.makedirs(new_directory)
 
 	for archived_site_path in old_directory.glob("*"):
-		shutil.move(str(archived_site_path), str(new_directory))
+		archived_site_path.rename(new_directory)
 
 	click.secho(f"Archived sites are now stored under {new_directory}")
 
--- a/bench/patches/v5/update_archived_sites.py
+++ b/bench/patches/v5/update_archived_sites.py
@@ -9,7 +9,6 @@ patch and try again later.
 Corresponding changes in frappe/frappe via https://github.com/frappe/frappe/pull/15060
 """
 import os
-import shutil
 from pathlib import Path
 
 import click
--- a/bench/patches/v5/update_archived_sites.py
+++ b/bench/patches/v5/update_archived_sites.py
@@ -40,7 +40,7 @@ def execute(bench_path):
 		os.makedirs(new_directory)
 
 	for archived_site_path in old_directory.glob("*"):
-		shutil.move(archived_site_path, new_directory)
+		shutil.move(str(archived_site_path), str(new_directory))
 
 	click.secho(f"Archived sites are now stored under {new_directory}")
 


bench/patches/v5/update_archived_sites.py
```diff
@@ -40,7 +40,7 @@ def execute(bench_path):
 		os.makedirs(new_directory)
 
 	for archived_site_path in old_directory.glob("*"):
-		shutil.move(archived_site_path, new_directory)
+		shutil.move(str(archived_site_path), str(new_directory))
```
gavindsouza: We could probably get rid of `shutil.move` altogether? 🤔 

```suggestion
		archived_site_path.rename(new_directory)
```

Ref: https://docs.python.org/3/library/pathlib.html#pathlib.Path.rename

rtdany10: Yes, that could be a better solution.

rtdany10: > Ref: https://docs.python.org/3/library/pathlib.html#pathlib.Path.rename

Wouldn't `replace` be a better than `rename`?
https://docs.python.org/3/library/pathlib.html#pathlib.Path.replace

gavindsouza: > Wouldn't `replace` be a better than `rename`?

I'm not sure how they differ. What I get from the documentation is that they are pretty much the same. If they are, I feel rename feels more natural to my mind than replace...given we may very rarely actually replace something.

rtdany10: > I'm not sure how they differ. What I get from the documentation is that they are pretty much the same.

The only difference I could find from the documentation:
Rename: **On Unix**, if target exists and **is a file**, it will be replaced silently if the user has permission.
Replace: If target points to an **existing file or directory**, it will be unconditionally replaced.

Suggested `replace` since we are dealing with directories.

gavindsouza: @rtdany10  Okay, I tried this with `.rename` and `.replace` with existing directories. Both failed the same way.

Screenshot 2021-12-20 at 2 36 46 PM

rtdany10: @gavindsouza I guess we can merge this PR then?

gavindsouza (approved):

PR #1239, opened 2021-12-20
feat: Inform user of change in configuration files

### LINKED ISSUE #1035: Watcher for config updates
Issue: Feature Request

There are updates in the bench config files and are usually security related. But they get overlooked as the configs don't get regenerated automatically. 

Something that checks if there are config changes that haven't been applied to your bench. It hints you to regenerate the configs again. This can be triggered on certain bench command executions similar to https://github.com/frappe/bench/pull/960. Move these to a watcher functions list that execute on triggers perhaps.

**Describe alternatives you've considered**
_None_

**Additional context**
_None_

anirudhdggl:
What type of a PR is this?

- [x] New Feature Submissions
- [ ] Bug Fix

---

## Explanation

Whenever there is a change in any config file in a bench, within the config directory, the user has to ensure that he restarts the bench. If at any time user forgets to restart the relevant bench, a warning will be presented to him on the CLI stating that he needs to restart the bench in order for the configuration changes to take effect

---

## Screenshots

image
image
image

closes: #1035

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,10 @@ Bench is a command-line utility that helps you to install, update, and manage mu
 	<a target="_blank" href="https://app.travis-ci.com/github/frappe/bench" title="CI Status">
 		<img src="https://app.travis-ci.com/frappe/bench.svg?branch=develop">
 	</a>
-	<a target="_blank">
+	<a target="_blank" href="https://pypi.org/project/frappe-bench" title="PyPI Version">
+		<img src="https://badge.fury.io/py/frappe-bench.svg" alt="PyPI version">
+	</a>
+	<a target="_blank" title="Platform Compatibility">
 		<img src="https://img.shields.io/badge/platform-linux%20%7C%20osx-blue">
 	</a>
 	<a target="_blank" href="https://app.fossa.com/projects/git%2Bgithub.com%2Ffrappe%2Fbench?ref=badge_shield" title="FOSSA Status">
--- a/bench/app.py
+++ b/bench/app.py
@@ -26,10 +26,7 @@ from bench.utils import (
 )
 from bench.utils.bench import (
 	build_assets,
-	get_env_cmd,
 	install_python_dev_dependencies,
-	restart_supervisor_processes,
-	restart_systemd_processes,
 )
 from bench.utils.render import step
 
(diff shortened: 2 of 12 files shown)


anirudhdggl:
@gavindsouza I'll be fixing the CI tests - but can you help me with the sematic PR?

I've mentioned the PR title as specified in the contribution guidelines and also included a commit for the same. Is something missing from the title / commit?

anirudhdggl:
Added a fix as well - for existing bench, where the config log filedo not exist. Initially this would've caused an error but now we would prompt the user to run `bench restart` command.

Bench restart would then go and create the config log file and from next time onwards whenever a change is made to any config file, there would be a message prompting users to restart bench

image

gavindsouza:
hey @anirudhdggl, thanks for your contribution. This is certainly going to be useful to remind users to reload their bench's processes. Although, one bit does come up that any command that updates config files also restarts the processes by default. This is a good-to-have but we could better move this elsewhere and not trigger it every time someone uses the CLI.

---

However, this doesn't resolve the mentioned issue. Issue #1035 is about telling users that they should reset their config files because we've made changes in them - this could be adding performance, security, or bug fixes. ie, whenever changes are made in the template files (ref:[nginx config template]( https://github.com/frappe/bench/blob/develop/bench/config/templates/nginx.conf)), and an end-user updates their bench package, there should be a mechanism in place that finds and resolves outdated configs.

anirudhdggl:
Okay - so basically #1035 is that whenever any file is changed in this [directory](https://github.com/frappe/bench/tree/develop/bench/config/templates) in GitHub, this exact location, we should ask the users to make the relevant changes to their local configs? Am I understanding it correctly?

And is this just for the `.conf` files or all the files (even within the subdirectory) in this directory?

gavindsouza:
> Okay - so basically #1035 is that whenever any file is changed in this [directory](https://github.com/frappe/bench/tree/develop/bench/config/templates) in GitHub, this exact location, we should ask the users to make the relevant changes to their local configs? Am I understanding it correctly?

Yes! Nicely put, thanks.



> And is this just for the `.conf` files or all the files (even within the subdirectory) in this directory?

This, I didn't fully follow.

---

@anirudhdggl Hope this thread didn't die out. I've been caught up with other things, apologies 🥲

PR #1241, opened 2021-12-22
fix: remove patches.txt as identifier for frappe app

ankush:
patches.txt will likely be removed in future https://github.com/frappe/frappe/pull/15351

In meantime, it's best to not consider this as an identifier for frappe app directory.

### DIFF
--- a/bench/utils/__init__.py
+++ b/bench/utils/__init__.py
@@ -20,7 +20,7 @@ from bench.exceptions import CommandFailedError, InvalidRemoteException, Validat
 
 logger = logging.getLogger(PROJECT_NAME)
 bench_cache_file = ".bench.cmd"
-paths_in_app = ("hooks.py", "modules.txt", "patches.txt")
+paths_in_app = ("hooks.py", "modules.txt")
 paths_in_bench = ("apps", "sites", "config", "logs", "config/pids")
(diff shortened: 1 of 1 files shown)


gavindsouza:
@ankush This feels like a very loose check; Only checking if `hooks.py` or `modules.txt` are in your directory to say "oh this is a frappe app". Any suggestions that can make this tighter? 🤔

ankush:
check for `patches.txt | patches.toml`?

patches.txt was anyway sorta problematic, it's not mandatory to have patches and if you remove file from your repo it's not a frappe app anymore 😅 

modules.txt and hooks.py are kinda mandatory... _for now_ 😅 

Maybe we can start creating a new file for frappe related meta info 😬 `frappe_app.toml` that will classify with 100% certainty.

gavindsouza:
> patches.txt was anyway sorta problematic, it's not mandatory to have patches and if you remove file from your repo it's not a frappe app anymore 😅

Who decides if it's mandatory or not? frappe/frappe and frappe/bench do. What's problematic is that the template generates everything that's needed but doesn't specify how critical each of the files are.

---

Just checking if a folder has `modules.txt` and a `hooks.py` to classify it as a frappe app seems like a very weak check to me now. Closing this. We can revisit this when required.

PR #1257, opened 2022-01-21
feat: Resolve and install

Aradhya-Tripathi:
### Dependency resolver

* Added `--resolve-deps` to get-app (get-app ++).
   * **Does not install dependencies of an app unless ```--resolve-deps``` flag is passed.**
   * Option first resolves all dependencies of the frappe app passed as an argument.
   * Dependencies read from hooks.py.
   * **Does not clone app repo before resolving** all of its dependencies and their dependencies.
   * After obtaining the resolution all the dependencies are installed.

* Added app states 
   * Added `app_states.json` to maintain app states i.e app versions, remotes, resolution, etc.

### DIFF
--- a/bench/app.py
+++ b/bench/app.py
@@ -1,4 +1,5 @@
 # imports - standard imports
+from collections import OrderedDict
 import functools
 import json
 import logging
@@ -46,6 +47,7 @@ class AppMeta:
 			3. frappe/healthcare@develop
 			4. healthcare
 			5. healthcare@develop, healthcare@v13.12.1
+			6. erpnext
 
 		References for Version Identifiers:
 		 * https://www.python.org/dev/peps/pep-0440/#version-specifiers
@@ -91,6 +93,7 @@ class AppMeta:
 			self._setup_details_from_name_tag()
 
 	def _setup_details_from_mounted_disk(self):
+		self.branch = subprocess.check_output(["git", "rev-parse", "--abbrev-ref", "HEAD"])
 		self.org, self.repo, self.tag = os.path.split(self.name)[-2:] + (self.branch,)
 
 	def _setup_details_from_name_tag(self):
@@ -98,6 +101,7 @@ class AppMeta:
 		self.tag = self.tag or self.branch
 
 	def _setup_details_from_installed_apps(self):
+		self.branch = subprocess.check_output(["git", "rev-parse", "--abbrev-ref", "HEAD"])
 		self.org, self.repo, self.tag = os.path.split(
 			os.path.join(self.bench.name, "apps", self.name)
 		)[-2:] + (self.branch,)
@@ -196,6 +200,35 @@ class App(AppMeta):
 	@step(title="Uninstalling App {repo}", success="App {repo} Uninstalled")
 	def uninstall(self):
 		self.bench.run(f"{self.bench.python} -m pip uninstall -y {self.repo}")
+
+	def _get_dependencies(self):
+		from bench.utils.app import get_required_deps_url
+		import toml
+
+		info_file = None
+		required_url = get_required_deps_url(self.url, self.tag)
+		print(required_url)
+		try:
+			info_file = toml.loads(requests.get(required_url).text)
+		except Exception:
+			click.echo("\nNo toml file found \n", err=True)
+
+		return info_file["required_apps"] if info_file  else {}
+
+def make_resolution_plan(app: App, bench):
+	"""
+	decide what apps and versions to install and in what order
+	"""
+	resolution = OrderedDict()
+	resolution[app.repo] = app
+	for app_name, branch in app._get_dependencies().items():
+		dep_app = App(app_name, branch=branch, bench=bench)
+		resolution[dep_app.repo] = dep_app
+		resolution.update(make_resolution_plan(dep_app, bench))
+		if app_name in resolution:
+			print("Resolve this conflict!")
+	return resolution
+
 
 
 def add_to_appstxt(app, bench_path="."):
@@ -310,6 +343,7 @@ def get_app(
 	repo_name = app.repo
 	branch = app.tag
 	bench_setup = False
+	apps_to_install = make_resolution_plan(app, bench)
 
 	if not is_bench_directory(bench_path):
 		if not init_bench:
@@ -320,8 +354,9 @@ def get_app(
 
 		from bench.utils.system import init
 
+		frappe_app = apps_to_install["frappe"]
 		bench_path = get_available_folder_name(f"{app.repo}-bench", bench_path)
-		init(path=bench_path, frappe_branch=branch)
+		init(path=bench_path, frappe_branch=frappe_app.tag)
 		os.chdir(bench_path)
 		bench_setup = True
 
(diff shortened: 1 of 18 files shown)


gavindsouza (commented):
After #1208, get-app resolves dependencies mentioned in required_apps and there's no turn off switch either. I realised that could be problematic given the resolution is still naive. Thought this PR would change that.

Ran the following two commands to check the difference between them. There was none. Both initialized bench with frappe@develop, erpnext@develop & healthcare@develop. 

* `bench --use-feature='dynamic-feed' get-app healthcare --init-bench`
* `bench --use-feature='dynamic-feed' get-app healthcare --init-bench --resolve-deps`

This behaviour makes sense though. With `--init-bench`, dependencies should always get resolved IMO. Else, your install is likely going to break.

---

But, when using get-app just to install apps on a particular bench. I would expect a difference between using `--resolve-deps` and not...In this case, there's not? 😅

Screenshot 2022-02-16 at 12 47 44 PM

Aradhya-Tripathi:
> After #1208, get-app resolves dependencies mentioned in required_apps and there's no turn off switch either. I realised that could be problematic given the resolution is still naive. Thought this PR would change that.
> 
> Ran the following two commands to check the difference between them. There was none. Both initialized bench with frappe@develop, erpnext@develop & healthcare@develop.
> 
> * `bench --use-feature='dynamic-feed' get-app healthcare --init-bench`
> * `bench --use-feature='dynamic-feed' get-app healthcare --init-bench --resolve-deps`
> 
> This behaviour makes sense though. With `--init-bench`, dependencies should always get resolved IMO. Else, your install is likely going to break.
> 
> But, when using get-app just to install apps on a particular bench. I would expect a difference between using `--resolve-deps` and not...In this case, there's not? sweat_smile
> 
> Screenshot 2022-02-16 at 12 47 44 PM

It would make more sense to use --resolve-deps once versioning is introduced, as of now the difference is internal, resolution is made internally before installation starts.

We could probably log the resolution or display it during --resolve-deps.

bench/app.py
```diff
@@ -171,42 +170,80 @@ def get(self):
...
+	@step(title="Cloning and installing {repo}", success="App {repo} Installed")
+	def install_resolved_apps(self, *args, **kwargs):
+		self.get()
+		self.install(*args, **kwargs, resolved=True)
+
 	@step(title="Uninstalling App {repo}", success="App {repo} Uninstalled")
 	def uninstall(self):
 		self.bench.run(f"{self.bench.python} -m pip uninstall -y {self.repo}")
 
+	def _get_dependencies(self):
```
gavindsouza, line 218: 1. Handle `on_disk`
2. Fix this

```python
>>> a = App("./apps/healthcare")
>>> a._get_dependencies()
  File "<stdin>", line 1, in <module>
  File "/Users/gavin/bench/bench/app.py", line 216, in _get_dependencies
    required_apps = eval(lines[0].strip("required_apps").strip().lstrip("=").strip())
IndexError: list index out of range
>>>
```

bench/app.py
```diff
@@ -310,16 +318,28 @@ def get_app(
...
 
 	bench = Bench(bench_path)
 	app = App(git_url, branch=branch, bench=bench)
 	git_url = app.url
 	repo_name = app.repo
 	branch = app.tag
 	bench_setup = False
+	frappe_path, frappe_branch = None, None
+
+	if resolve_deps:
```
gavindsouza, line 346: Make behaviour consistent for command with and without `--resolve-deps`

bench/app.py
```diff
@@ -371,22 +405,75 @@ def get_app(
...
+	for repo_name, app in reversed(resolution.items()):
+		existing_dir, path_to_app = check_existing_dir(bench_path, repo_name)
+		if existing_dir:
+			is_compatible = False
+
+			try:
+				installed_branch = bench.apps.states[repo_name]["resolution"]["branch"].strip()
+			except:
+				installed_branch = (
+					subprocess.
```
gavindsouza, line 442: Check scope for using the `gitpython` library here

bench/utils/app.py
```diff
@@ -107,7 +107,9 @@ def switch_to_develop(apps=None, bench_path=".", upgrade=True):
 
 
 def get_version_from_string(contents, field="__version__"):
-	match = re.search(r"^(\s*%s\s*=\s*['\\\"])(.+?)(['\"])(?sm)" % field, contents)
+	match = re.search(r"^(\s*%s\s*=\s*['\\\"])(.+?)(['\"])" % field, contents, flags=(re.S | re.M))
+	if not match:
+		raise Exception("No match was found")
```
gavindsouza: Throw named exceptions - also sonarcloud

bench/utils/__init__.py
```diff
@@ -48,6 +49,37 @@ def is_frappe_app(directory: str) -> bool:
...
+	:param frappe_branch: branch to check
+	:type frappe_branch: str
+	:raises InvalidRemoteException: branch for this repo doesn't exist
+	"""
+	if "http" in frappe_path and frappe_branch:
+		frappe_path = frappe_path.replace(".git", "")
+		try:
+			owner, repo = frappe_path.split("/")[3], frappe_path.split("/")[4]
+		except IndexError:
+			raise InvalidRemoteException
```
gavindsouza: Add a message for this? Seems confusing when hit with tracebacks (ref: https://github.com/frappe/bench/issues/1267)

bench/app.py
```diff
@@ -171,42 +170,80 @@ def get(self):
...
-		setup_app_dependencies(
-			repo_name=self.repo,
-			bench_path=self.bench.name,
-			branch=self.tag,
-			verbose=verbose,
-			skip_assets=skip_assets,
-		)
+		if not resolved and self.repo != "frappe":
+			click.secho(
+				f"Ignoring dependencies of {self.name} to install dependencies use --resolve-deps",
```
gavindsouza: Handle this bit with the setup requirements command too. Perhaps add a `--resolve-deps` option too and don't show this message when called from there anyway

gavindsouza (changes requested):

bench/app.py
Aradhya-Tripathi: This is just to provide support for frappe versions (when that's mentioned in the hooks of the required apps).

bench/app.py
Aradhya-Tripathi: Not showing the message when called from setup requirements.

bench/utils/__init__.py
Aradhya-Tripathi: Resolved in [cd583f7](https://github.com/frappe/bench/pull/1257/commits/cd583f7822a5aaa29f09cc9ba016b6b23c1d9cc2)

bench/utils/app.py
Aradhya-Tripathi: Resolved in 
a32ceb4

bench/app.py
Aradhya-Tripathi: fixed by [0b862c0](https://github.com/frappe/bench/pull/1257/commits/0b862c07457c9914840bd0ce87a90306fec37456)

bench/app.py
Aradhya-Tripathi: Git gives a lot more flexibility with ease so I think it would be better to use git directly here?

gavindsouza (approved):
Works as expected, LGTM

PR #1281, opened 2022-03-17
fix: setup supervisor

gavindsouza:
Slightly better handling for errors on servers with much older setups. Also, trying to fix [CI failure](https://app.travis-ci.com/github/frappe/bench/jobs/563662216) and issues (https://discuss.erpnext.com/t/unable-to-initialize-bench/87449) caused by 25f49c2dbba998a7b11975a6f17034bdb050c3ab

### DIFF
--- a/bench/app.py
+++ b/bench/app.py
@@ -320,6 +320,7 @@ def get_app(
 	repo_name = app.repo
 	branch = app.tag
 	bench_setup = False
+	restart_bench = not init_bench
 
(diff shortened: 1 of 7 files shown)


gavindsouza:
~~> Also, trying to fix [CI failure](https://app.travis-ci.com/github/frappe/bench/jobs/563662216)~~

~~Seems like it may be unrelated since #1257's CI is also failing~~

Okay, it's related to #1280

PR #1265, opened 2022-02-12
feat: Setup Requirements for specific apps

DaizyModi:
Added `apps` argument to run setup requirements for a specific app(s)

## Usage

Specify multiple space-separated apps. Similar to how `bench switch-to-branch` works.
**CAVEAT**: apps must be specified in the end.
```bash
bench setup requirements [OPTIONS] [APPS]
```

## Other Changes
- Don't restart bench or build assets while setting up dependencies (original behavior). Restart and build should be separately executed. Restart fails for developer setups anyway:

```
ERROR: Command 'supervisorctl status' returned non-zero exit status 127.
Traceback (most recent call last):
  File "/usr/local/bin/bench", line 33, in <module>
    sys.exit(load_entry_point('frappe-bench', 'console_scripts', 'bench')())
  File "/home/snv/Work/.bench/bench/cli.py", line 121, in cli
    raise e
  File "/home/snv/Work/.bench/bench/cli.py", line 111, in cli
    bench_command()
  File "/home/snv/.local/lib/python3.8/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/home/snv/.local/lib/python3.8/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/home/snv/.local/lib/python3.8/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/snv/.local/lib/python3.8/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/snv/.local/lib/python3.8/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/snv/.local/lib/python3.8/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/home/snv/Work/.bench/bench/commands/setup.py", line 150, in setup_requirements
    bench.setup.requirements(apps=apps)
  File "/home/snv/Work/.bench/bench/utils/render.py", line 95, in wrapper_fn
    return fn(*args, **kwargs)
  File "/home/snv/Work/.bench/bench/bench.py", line 342, in requirements
    App(app, bench=self.bench, to_clone=False).install()
  File "/home/snv/Work/.bench/bench/utils/render.py", line 110, in wrapper_fn
    return fn(*args, **kwargs)
[...]

PR #1285, opened 2022-03-22
fix: setup all apps is empty list is supplied

ankush:
`bench setup requirements --dev` doesn't do anything because `apps` is `()`

caused by https://github.com/frappe/bench/pull/1265 which changes apps from `None` to `()`


gavindsouza (approved):

ankush:
Screenshot 2022-03-22 at 1 00 41 PM


shallow scan for similar problems. looks good.

gavindsouza:
Travis is choking. Will merge for now given the breaking CI builds that depend on this 🥲

PR #1291, opened 2022-04-06
fix: Check for unauthorised status code in response

ChillarAnand:
Github is returning 403 status on some cases. 

Screenshot 2022-04-06 at 12 04 40

### DIFF
--- a/bench/utils/__init__.py
+++ b/bench/utils/__init__.py
@@ -429,7 +429,7 @@ def find_org(org_repo):
 
 	for org in ["frappe", "erpnext"]:
 		res = requests.head(f"https://api.github.com/repos/{org}/{org_repo}")
-		if res.status_code == 400:
+		if res.status_code in (400, 403):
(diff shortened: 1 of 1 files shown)


bench/utils/__init__.py
```diff
@@ -429,7 +429,7 @@ def find_org(org_repo):
 
 	for org in ["frappe", "erpnext"]:
 		res = requests.head(f"https://api.github.com/repos/{org}/{org_repo}")
-		if res.status_code == 400:
+		if res.status_code in (400, 403):
 			res = requests.head(f"https://github.com/{org}/{org_repo}")
```
Aradhya-Tripathi, line 433: I think if the API returns 403 then the open route won't work either 🤔 
There could be a separate exception for it for the time being

ChillarAnand: Nope. For the other URL, it is working fine.

Screenshot 2022-04-08 at 14 28 55

Only for api.github.com, it is returning 403.

gavindsouza (approved):

PR #1306, opened 2022-05-15
fix: allow bench init with git tags

revant:
```
bench init --skip-redis-config-generation --frappe-branch v13.14.1 frappe-bench
```

Results into:

```
bench.exceptions.InvalidRemoteException: Invalid v13.14.1 for the remote https://github.com/frappe/frappe.git
```

Fix allows git tags to be used for `--frappe-branch` during `bench init`

### DIFF
--- a/bench/utils/__init__.py
+++ b/bench/utils/__init__.py
@@ -61,35 +61,19 @@ def is_valid_frappe_branch(frappe_path:str, frappe_branch:str):
 	:type frappe_branch: str
 	:raises InvalidRemoteException: branch for this repo doesn't exist
 	"""
-	import subprocess
+	import git
+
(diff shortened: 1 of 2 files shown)


bench/utils/__init__.py
```diff
@@ -50,6 +50,7 @@ def is_frappe_app(directory: str) -> bool:
 	return bool(is_frappe_app)
 
 
+@lru_cache(maxsize=None)
```
phot0n, line 53: PS: cache is being used here so that when the same function is called multiple times in the same process, it doesn't need to ask for git creds multiple times if http(s) is being used

phot0n (approved):
LGTM 👍

PR #1312, opened 2022-05-24
feat(config): Add IPv6 listener to nginx site configuration

### LINKED ISSUE #814: ipv6 nginx config
I was struggling with an instance where I kept receiving the default web page instead of my erpnext instance. 
After debugging I noticed that in 
`/home/frappe/frappe-bench/config/nginx.conf`

bench setup nginx command doesn't include ipv6 listen directive. 

It generates the file like this

```
server {
        listen 80;
        server_name
                <domain.com>
                ;

        root /home/frappe/frappe-bench/sites;
```

In order towork with ipv6 as well it needs to create it like this

```
server {
        listen 80;
        listen [::]:80;
        server_name
                <domain.com>
                ;

        root /home/frappe/frappe-bench/sites;
```

deajan:
closes #814 

What type of a PR is this?

- [X] Changes to Existing Features

---

- [X] Have you followed the guidelines in our Contributing document?
- [X] Have you checked to ensure there aren't other open [Pull Requests](../pulls) for the same update/change?
- [X] Have you lint your code locally prior to submission?
- [X] Have you successfully run tests with your changes locally?
- [X] Does your commit message have an explanation for your changes and why you'd like us to include them?
- [N/A] Docs have been added / updated
- [N/A] Tests for the changes have been added (for bug fixes / features)
- [N/A] Did you modify the existing test cases? If yes, why?

---

This PR adds an IPv6 listener to nginx configuration files.
With RHEL 8, localhost TCP calls are done via IPv6 if available.
Thus, even when configuring `/etc/hosts` file to resolve a bench site name to 127.0.0.1 and ::1, a command like `curl localhost` will read nginx default site instead of the frappe configured site.
This also resolves an issue where `wkhtmltodpf` misses CSS and images since it will try to load assets on the default nginx site.

### DIFF
--- a/bench/config/templates/nginx.conf
+++ b/bench/config/templates/nginx.conf
@@ -14,8 +14,10 @@ map {{ from_variable }} {{ to_variable }} {
 server {
 	{% if ssl_certificate and ssl_certificate_key %}
 	listen {{ port }} ssl;
+	listen [::]:{{ port }} ssl;
 	{% else %}
 	listen {{ port }};
+	listen [::]:{{ port }};
 	{% endif %}
 
 	server_name


gavindsouza (approved):

iptelephony:
This creates a problem for hosts that have ipv6 disabled. `bench setup production` will fail at `nginx -t`. A better approach would be to only add the ipv6 listener if and only if the host has ipv6 enabled.

PR #1313, opened 2022-05-25
refactor(bench utils): migrated venv creation to python's builtin venv

zardini123:
What type of a PR is this?

- [x] Changes to Existing Features
- [ ] New Feature Submissions
- [ ] Bug Fix

---

The following pull request removes the usage of `virutalenv` for Python's built in `venv`.  This change has the most noticeable effect on `branch init`.

## Motivation

The development of this change was initiated due to the newest version of `virtualenv` being unable to produce a `bin/` folder on my server's linux distribution.  In contrast, python's built in `venv` (i.e. `python3 -m venv env`) properly created the virtual environment.

By making this PR, I'd like to make the argument that moving bench to `venv` helps reduce dependencies required for bench, which makes bench more accessible.  `venv` has been in the Python standard module's since version 3.3 (see [PEP 405](https://peps.python.org/pep-0405/)).  As bench is supported for versions 3.6, removing the dependency of virtualenv does not affect on compatibility.

## Changes

The prominent change this PR makes is removing `get_venv_path()` and adding `create_venv()` to the API of bench/utils/bench.py.

This change fixes an issue regarding separation of concerns with `get_venv_path()`.  There was fallback code in `get_venv_path()` if there was no `virtualenv` is present in `$PATH` by returning `python3 -m venv`.  This fallback proved useless, as the result of `get_venv_path()` being fed into command arguments in `env()` evaluated to `python3 -m venv -qq env -p python3`, which is not a valid command for `python3`, and results in a failure.

## Further Considerations

If `virtualenv` is still of use, I'd recommend bench to modify the CLI to provide user control of choosing between `venv`, `virtualenv`, or an external virtual environment manager (for example a virtual environment manager that also supports `conda`).  As provided above, auto-switching between the two can be difficult for maintenance.  I am very much open to having a conversation regarding this consideration.


gavindsouza:
Hey @zardini123, I've made some changes addressing similar issues via #1314. I've kept venv as a fallback and used virtualenv in case it's installed on the system. Do you propose any more changes to that?

---

PS: I was setting up a new machine and since virtualenv wasn't in PATH, things started to fall apart and fixed it via the linked PR.

zardini123:
Hi @gavindsouza, thank you for your changes!  And thank you for asking for my input!

I believe that preserving the getting of the executable path methodology is resulting in a lot of code bloat.  Currently there are two places in bench where venv's are created: one in `BenchSetup`, and one in `migrate_env`.  As of currently, `migrate_env` in your commit is not fixed by the changes, as the issue I mentioned above of mix of command-line arguments still holds.  I'd imagine copying and pasting of code from `BenchSetup` would be required to fix `migrate_env` due to preserving the get executable path method.

My changes proposes an API change that just simply makes a `create_env` function, and any crazy or interesting virtualenv path recognition can go in there.  It would help condense the source code in my opinion.

Curious to hear what you think!

gavindsouza:
@zardini123 Will you make the changes you're proposing in this PR?

zardini123:
Hi @gavindsouza, if you believe the changes are worth while, I'll resolve my merge conflicts and make the changes available later today!

gavindsouza:
I'm all for de-duplication. The linked PR uses the `virtualenv` wrapper if it exists else falls back to venv. I think we can get rid of virtualenv altogether and just keep it simple.

zardini123:
Attempting to rebase my fork resulted in the PR auto-closing.  Should be fixed now.

Issue #1174, opened 2021-05-13
Monorepo management


revant:
get-monorepo `bench get-monorepo <repo-url> <local-dir>`:

- clone `repo-url` into `local-dir`
- sym-link `apps` from `local-dir` to `frappe-bench/apps`
- install apps from into `frappe-bench/env`
- store info about install in `frappe-bench/monorepo.json`

Add `bench install-monorepo local-dir` based on above 

Monorepo Structure:

```
- apps/
  - app1
  - app2
  - appN
- frappe_apps.json
```

frappe_apps.json (optional, used for release management and advance tools):

```json
{
  "apps": [
    {
      "name": "app1", // exact folder name of app, don't allow apps of same name in a bench during sync.
      "version": "1.0.1"
    },
    {
      "name": "app2",
      "version": "1.3.1"
    }
  ]
}
```

Other tools:

- Release version for single app, tags: app1@1.1.1, app2@1.0.1
- bench sync-monorepo
- bench remove-monorepo
- https://github.com/frappe/bench/issues/1173

revant:
Execute from `~/frappe-bench` directory

`/path/to/script/bench_sync_monorepo.py --help`

`/path/to/script/bench_sync_monorepo.py ~/path/to/monorepo --force --verbose`

```python
#!/usr/bin/env python3

import argparse
import os
import subprocess
import sys


def main():
	parser = get_args_parser()

	args = parser.parse_args()

	if len(sys.argv) == 1:
		parser.print_help(sys.stderr)
		exit(1)

	apps = []
	apps_path = f"{args.directory}/apps"
	try:
		for subdir in os.listdir(apps_path):
			apps.append({"directory": os.path.join(apps_path, subdir), "app_name": subdir})

	except FileNotFoundError as e:
		handle_error(e)

	if len(apps) > 0:
		for app in apps:
			app_name = app.get("app_name")
			symlink_force(
				app.get("directory"), os.path.join(os.getcwd(), "apps", app_name), args.force,
			)

			quiet_flag = "" if args.verbose else "-q"
			shell_command = f"./env/bin/pip install {quiet_flag} -e apps/{app_name}"
			print("\033[1m" + shell_command + "\033[0m")
			subprocess.call(shell_command, shell=True)

		append_apps_txt(apps)

		shell_command = "bench setup requirements"
		print("\033[1m" + shell_command + "\033[0m")
		subprocess.call(shell_command, shell=True)

		shell_command = "bench build"
[...]

PR #1315, opened 2022-06-03
feat: get-app support soft link as alternative to git clone

dj12djdjs:
# Use Case
I'm experimenting with GitHub Codespaces (Only been using it for a couple of hours) So maybe I'm taking the wrong approach to this.

Let's say I'm working on `erpnext`, I'll create a code space for my fork `dj12djdjs/erpnext` which will put my working directory at `/workspaces/erpnext`. I can't test my code just like this. I need frappe and bench utilities to actual use ERPnext. so in the parent directory of workspaces I'll create `/frappe-bench` with frappe and the python environment. 

If I want to test my code, I'll have to commit to the github repo -> push -> then pull in the /frappe-bench/apps/erpnext

It would be much better if the changes I made in `/workspaces/erpnext` directly reflect to `/frappe-bench/apps/erpnext`

# Proposed Solution
Introduce the `--soft-link` flag to `get-app` command, altering the way bench will get the app into the apps directory. Instead of `git clone`,  it will `ln -s` if the `--soft-link` flag is provided.
get app ln-s

**Note:** I just through this together to test it, It broke some things like `bench remove-app` and maybe some others. I just wanted to get some feedback if others might find this a useful addition.

### DIFF
--- a/bench/app.py
+++ b/bench/app.py
@@ -13,6 +13,8 @@ from datetime import date
 from urllib.parse import urlparse
 import os
 
+from markupsafe import soft_str
+
 # imports - third party imports
 import click
 from git import Repo
@@ -158,8 +160,9 @@ class AppMeta:
 
 @functools.lru_cache(maxsize=None)
 class App(AppMeta):
-	def __init__(self, name: str, branch: str = None, bench: "Bench" = None, *args, **kwargs):
+	def __init__(self, name: str, branch: str = None, bench: "Bench" = None, soft_link : bool = False, *args, **kwargs):
 		self.bench = bench
+		self.soft_link = soft_link
 		self.required_by = None
 		self.local_resolution = []
 		super().__init__(name, branch, *args, **kwargs)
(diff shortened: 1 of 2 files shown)


gavindsouza (approved):

ankush:
I think this might have broken ERPNext CI

it's installing v14 instead somehow on v13 CI 👀 


https://github.com/frappe/erpnext/runs/6793355216?check_suite_focus=true 


Edit: ignore, I am not so sure about this after looking at diff.

dj12djdjs:
Okay, I'm surprised this was merged. I had mentioned it does break the command `bench remove-app` and that I haven't done much testing.

>Note: I just through this together to test it, It broke some things like bench remove-app and maybe some others. I just wanted to get some feedback if others might find this a useful addition.

Since, it's already merged, I'll do some more tests today and create a PR for fixes to remove-app and any others I find.

dj12djdjs:
Disregard, it is working fine. Somehow, the first time I ran bench remove-app it failed. I'm unable to replicate that now.

phot0n:
does this implement this? https://github.com/frappe/bench/issues/1174

from a glance looks like it :eyes:

PR #1327, opened 2022-06-28
refactor: using venv for virtual environment creation

Aradhya-Tripathi:
* Refactored bench setup and migrate-env to use venv.

Wrapping Up -> #1313

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,5 +5,5 @@ Jinja2~=3.0.3
 python-crontab~=2.4.0
 requests
 semantic-version~=2.8.2
-setuptools
+setuptools==44.0.0
 tomli;python_version<"3.11"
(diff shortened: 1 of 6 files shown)


requirements.txt
```diff
@@ -5,5 +5,5 @@ Jinja2~=3.0.3
 python-crontab~=2.4.0
 requests
 semantic-version~=2.8.2
-setuptools
+setuptools==44.0.0
```
gavindsouza, line 8: Hard pins are a bad idea for globally installed tools

```suggestion
setuptools>44.0.0
```

gavindsouza (changes requested):
Is this ready? And is this an alternative to https://github.com/frappe/bench/pull/1313?

requirements.txt
Aradhya-Tripathi: Not sure this isn't a fix it's still failing on install thought it was a setup tools versioning issue as it turns out it wasn't.

PR #1332, opened 2022-07-13
fix: Install gunicorn for production setup

ChillarAnand:
Related to https://github.com/frappe/frappe/pull/17482

Currently, gunicorn is added as a dependency for frappe. However gunicorn is not needed for non-production(dev, testing, CI etc) setups. 

Moving gunicorn installation to `bench setup production`.

### DIFF
--- a/bench/config/production_setup.py
+++ b/bench/config/production_setup.py
@@ -17,9 +17,11 @@ logger = logging.getLogger(bench.PROJECT_NAME)
 
 
 def setup_production_prerequisites():
-	"""Installs ansible, fail2banc, NGINX and supervisor"""
+	"""Installs ansible, gunicorn, fail2banc, NGINX and supervisor"""
 	if not which("ansible"):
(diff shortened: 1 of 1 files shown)


bench/config/production_setup.py
```diff
@@ -17,9 +17,11 @@
 
 
 def setup_production_prerequisites():
-	"""Installs ansible, fail2banc, NGINX and supervisor"""
+	"""Installs ansible, gunicorn, fail2banc, NGINX and supervisor"""
 	if not which("ansible"):
 		exec_cmd(f"sudo {sys.executable} -m pip install ansible")
+	if not which("gunicorn"):
```
gavindsouza, line 23: Don't think you're going to find `gunicorn` in PATH. You'll have to check if gunicorn exists in env manually. Unless, I'm missing something?

gavindsouza (changes requested):

PR #1333, opened 2022-07-25
fix(nginx)!: set `proxy_http_version` to `1.1`

lhjt:
What type of a PR is this?

- [ ] New Feature Submissions
- [x] Bug Fix

---

This sets the `proxy_http_version` to `1.1` within the `@webserver` block.

---

The reason this change should be implemented is that when this is deployed in a Kubernetes environment with Istio, the proxy request fails with HTTP error 426 as the HTTP version is too low for envoy to handle. Setting it to `1.1` will solve the issue.

---

> Screenshots/GIFs

image

### DIFF
--- a/bench/config/templates/nginx.conf
+++ b/bench/config/templates/nginx.conf
@@ -91,6 +91,7 @@ server {
 	}
 
 	location @webserver {
+		proxy_http_version 1.1;
 		proxy_set_header X-Forwarded-For $remote_addr;
 		proxy_set_header X-Forwarded-Proto $scheme;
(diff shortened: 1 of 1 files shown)


lhjt:
A temporary fix is to enable `http/1.0` support between sidecars in istio:

```yaml
apiVersion: install.istio.io/v1alpha1
kind: IstioOperator
spec:
  values:
    pilot:
      env:
        PILOT_HTTP10: '1'
```

gavindsouza (approved):

gavindsouza:
For reference: https://stackoverflow.com/questions/246859/http-1-0-vs-1-1

PR #1344, opened 2022-07-29
fix: Use subscript instead of method for regex

gavindsouza:
This is another dumb attempt to trigger a release because it doesn't get triggered without PRs linked or even if they're refactors!

Ref: 
- #1343 
- https://github.com/frappe/bench/commit/147aab1e4ecd9445135799e48829488c65b21e8e
- https://github.com/frappe/bench/commit/a0a81330326f5268790ed9246857d5e2139a2143
- https://github.com/frappe/bench/actions/runs/2762911056
- https://github.com/frappe/bench/actions/runs/2762866297

### DIFF
--- a/bench/utils/app.py
+++ b/bench/utils/app.py
@@ -250,7 +250,7 @@ def get_app_name(bench_path: str, folder_name: str) -> str:
 	if not app_name:
 		# retrieve app name from setup.py as fallback
 		with open(setup_py_path, "rb") as f:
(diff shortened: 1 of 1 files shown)


gavindsouza:
Refs mentioned in the description are attempts to fix packaging issue introduced via https://github.com/frappe/bench/pull/1337

https://discuss.erpnext.com/t/caution-somebody-broke-bench-last-night/92647/2

PR #1345, opened 2022-07-31
fix!: Remove backup command

gavindsouza:
bench's backup command is redundant and acts as a wrapper over Frappe's
backup command. This caused a conflict with Frappe's CLI after the
resolution correction introduced via a6f196440ad8c9b65a5bbc31a88d8c80f2ba2ded

ref: https://github.com/frappe/frappe/runs/7592331617?check_suite_focus=true

### DIFF
--- a/bench/commands/__init__.py
+++ b/bench/commands/__init__.py
@@ -73,7 +73,6 @@ bench_command.add_command(switch_to_develop)
 
 from bench.commands.utils import (
 	backup_all_sites,
-	backup_site,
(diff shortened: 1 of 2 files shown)


ankush (approved):

ankush:
PS: Avoiding going into the "override" rabbit hole. To avoid the possibility of another `hooks` mess where resolution and priority are ill-defined. Really doesn't feel necessary here.

PR #1362, opened 2022-08-23
fix: use ref param instead of branch for get_required_deps (GitHub contents API)

FHenry:
https://api.github.com/repositories/frappe/contents/erpnext/hooks.py?branch=version-13

return hook.py from develop branch 

https://api.github.com/repositories/frappe/contents/erpnext/hooks.py?ref=version-13

return hook.py from version-13 branch 

This cause installation of payment app for version-13 erpnext that is not use in version-13

https://docs.github.com/en/rest/repos/contents

```
Query parameters
Name, Type, Description
ref string

The name of the commit/branch/tag. Default: the repository’s default branch (usually master)
```

### DIFF
--- a/bench/utils/app.py
+++ b/bench/utils/app.py
@@ -118,7 +118,7 @@ def get_version_from_string(contents, field="__version__"):
 	)
 	if not match:
 		raise VersionNotFound(f"{contents} is not a valid version")
-	return match.grou,000p(2)
+	return match.group(2)
 
 
 def get_major_version(version):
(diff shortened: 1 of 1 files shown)


ankush:
As per docs looks like `branch` is only supported in body, since we are doing it using query params `ref` is required :+1: 

Easy to verify using download link suggested in API response: 


https://api.github.com/repos/frappe/erpnext/contents/erpnext/hooks.py?branch=version-13 this points to develop

https://api.github.com/repos/frappe/erpnext/contents/erpnext/hooks.py?ref=version-13 points to correct branch

ankush (approved):

PR #1364, opened 2022-08-26
fix: parsing error in sudoers file

### LINKED ISSUE #1320: Error in sudoers file created by bench
```
fatal: [localhost]: FAILED! => {
    "changed": false,
    "module_stderr": ">>> /etc/sudoers.d/frappe: syntax error near line 19 <<<\nsudo: parse error in /etc/sudoers.d/frappe near line 19\nsudo: no valid sudoers sources found, quitting\nsudo: unable to initialize policy plugin\n",
    "module_stdout": "",
    "msg": "MODULE FAILURE\nSee stdout/stderr for the exact error",
    "rc": 1
}
```

rtdany10:
Closes #1320

### DIFF
--- a/bench/config/templates/frappe_sudoers
+++ b/bench/config/templates/frappe_sudoers
@@ -17,3 +17,4 @@
 
 {{ user }} ALL = (root) NOPASSWD: {{ certbot }}
 Defaults:{{ user }} !requiretty
+


ankush:
@rtdany10 can you give me some more context on how newline solves parsing error? :thinking: 


Especially because the file already has one extra newline (typical newline on EOF present in all files), GitHub just doesn't render it because it's common practice.

ankush (approved):
The template needs an extra newline because the rendered file doesn't contain any right now.

PR #1372, opened 2022-09-17
fix: shutil.rmtree does not support symlink

dj12djdjs:
**Introduced In:**  https://github.com/frappe/bench/pull/1315

https://docs.python.org/3/library/shutil.html#shutil.rmtree
> Delete an entire directory tree; path must point to a directory **(but not a symbolic link to a directory).**

### DIFF
--- a/bench/app.py
+++ b/bench/app.py
@@ -201,7 +201,10 @@ class App(AppMeta):
 		active_app_path = os.path.join("apps", self.name)
 
 		if no_backup:
-			shutil.rmtree(active_app_path)
+			if not os.path.islink(active_app_path):
+				shutil.rmtree(active_app_path)
+			else:
+				os.remove(active_app_path)
 			log(f"App deleted from {active_app_path}")
 		else:
 			archived_path = os.path.join("archived", "apps")


bench/app.py
```diff
@@ -201,7 +201,10 @@ def remove(self, no_backup: bool = False):
 		active_app_path = os.path.join("apps", self.name)
 
 		if no_backup:
-			shutil.rmtree(active_app_path)
+			if not os.path.islink(active_app_path):
+				shutil.rmtree(active_app_path)
+			else:
+				os.remove(active_app_path)
```
ankush, line 207: Errrm... os.remove doesn't work with folders? 

check this: https://docs.python.org/3/library/os.html#os.remove

ankush (changes requested):

bench/app.py
dj12djdjs: @ankush To my understanding a symbolic link is a file. The file is just a pointer to ether another file or directory.

I'll double check the execution of this, and attach a screen capture to prove its case.

ankush: My bad. This makes sense 👍

ankush (approved):

PR #1382, opened 2022-11-02
refactor: Drop redis_socketio service

ankush:
Summary:
- After this change bench wont generate config for redis_socketio
  instance.
- Instead we set the same port as redis_cache service for use in
  socketio. i.e. both will share same instance.

Code changes:

- Config will have this difference:

```diff
- "redis_socketio": "redis://localhost:12000",
- "redis_queue": "redis://localhost:11000",
+ "redis_socketio": "redis://localhost:11000",
+ "redis_queue": "redis://localhost:11000",
```

- supervisord/systemd wont start redis_socketio service
- for development setups: redis_socketio service is removed from procfile.

Why?

- Currently this redis instance is used only to ship messages from python
to nodejs server... which is largely stateless (only pub/sub channels)
- This change reduces 1 running process. Less resources for this ~= more
  resources for other running things.

All dependent deployment projects will likely have to refactor their code to drop this service, shouldn't be too much work.
Example: Frappe cloud (already done), Frappe Docker.


ankush:
not finding time to finish it, also 5-10MB memory saving at max :pensive:  


You can achieve this in your deployments just by doing the config change that I mentioned in description.

SaiFi0102:
Good idea, but could have been backwards compatible

ankush:
@SaiFi0102 this is backward compatible AFAIK, unless you need 3rd service running for something else?

SaiFi0102:
I did a `bench setup supervisor` after migrating and old instance to v14. Since `common_site_config.json` was configured for port 12000... the bench got stuck until I found out about this change. So I'm guessing others can also fall into this issue.

ankush:
@SaiFi0102 that should be fixable by patching common_site_config when setting up supervisor :+1:

ankush:
@SaiFi0102 https://github.com/frappe/bench/pull/1481 this should fix it.

PR #1387, opened 2022-11-22
feat: add `max_requests` to gunicorn args

ankush:
As gunicorn is long running process, potentially running for days
without restart the workers might start accumulating garbage that's
never cleaned up and memory usage spikes after some use.

This largely happens because of third-party module imports like pandas,
openpyxl, numpy etc. All of these are required only for few requests and
can be easily re-loaded when required.

`max_requests` restarts the worker after processing number of configured
requests.

How to use?
- If you have more than 1 gunicorn workers then this is automatically
  enabled. You can tweak the max_requests parameter with
  `gunicorn_max_requests` key in common_site_config
- If you just have 1 gunicorn worker (not recommended) then this is not
  automatically enabled as restarting the only worker can cause spikes
  in response times whenever restart is triggered. You can still enable by
  setting same config key.


ankush:
Note: Prod test is failing silently.

ankush:
- [ ] docs
- [ ] fixup prod tests

PR #1391, opened 2022-11-24
fix: restart proc manager if set in config

ankush:
This used to be the behaviour before, looks like it was changed unintentionally.

https://github.com/frappe/bench/blob/cbf30b4e6a77132bb242606e1d3b3367fab4f3ee/bench/utils/bench.py#L417

Current behaviour doesn't make much sense. It requires passing a flag + setting config both to trigger restart.

alt to https://github.com/frappe/bench/pull/1390

### DIFF
--- a/bench/bench.py
+++ b/bench/bench.py
@@ -147,7 +147,7 @@ class Bench(Base, Validator):
 
 		if conf.get("developer_mode"):
 			restart_process_manager(bench_path=self.name, web_workers=web)
-		if supervisor and conf.get("restart_supervisor_on_update"):
+		if supervisor or conf.get("restart_supervisor_on_update"):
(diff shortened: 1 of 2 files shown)


ankush:
no idea why tests are failing on systemd restart here :roll_eyes:

arvind-hasti:
@ankush 
Hey, I wanna suggest changing this to "and". (only a suggestion, I'm not very familiar with the repo)
In my case, bench update performs its updates but exits in error because the supervisor is not working.
Changing the "restart_supervisor_on_update" config to False has no effect. Config is not affecting the flow then whats the use of it here?
Because bench update exits in error even tho the actual updates are successful, this makes the website stuck in the "HTTP 503 Updating website, come back later" page.
Had to down grade bench to 5.14 to get over this.

PR #1396, opened 2022-11-29
fix: migrate env

ankush:
- dont fail silently
- Use correct path for python exec (`bench/env` -> `bench/env/bin/python`)
- Install frappe first and re-install one app at time.
- Use `venv` instead of `virtualenv`
- Don't attmept to migrate if target is present in current venv (LOL?)

Tested with.

- 3.7 -> 3.10
- 3.10 -> 3.11
- 3.11 -> 3.10


sagarvora:
Thanks this is really helpful. It used to be my go-to command for migrating, before it broke.

PR #1397, opened 2022-12-01
feat: New™️ easy-install.py

athul:
This PR includes
- New containerized setup of Frappe and ERPNext
-  Uses frappe_docker
- Generates compose files on the fly and spins up a new production server
- Installs and sets up docker if not present
- Made with standard library
- Can be used for development and production setup

Special thanks to @revant for helping me out on this <3

closes https://github.com/frappe/frappe/issues/18223

---

TODO:
- [x] Resolve open issues
- [x] Manual tests @ankush / @revant 
- [x] Automated test for script on github action @athul 
- [x] Add "copy and run in shell" 2-3 liners in README.md of bench and frappe

### DIFF
--- a/easy-install.py
+++ b/easy-install.py
@@ -1,0 +1,239 @@
+#!/usr/bin/env python3
+import argparse
+import subprocess
+import os
+import sys
+import time
+from shutil import which
+import platform
+from secrets import token_bytes
+from base64 import b64encode
+
+CRED = "\033[31m"
+CEND = "\033[0m"
+CGRN = "\33[92m"
+CYLW = "\33[93m"
+
+
+def clone_frappe_docker_repo() -> None:
+    try:
+        subprocess.run(
+            ["git", "clone", "https://github.com/frappe/frappe_docker"], check=True
+        )
+    except Exception as e:
+        print(f"\n{CRED}Cloning frappe_docker Failed{CEND}\n\n", e)
+
+
+def write_to_env(wd, site,db_pass,email):
+    site_name = site if site != "" else ""
+    with open(os.path.join(wd, ".env"), "w") as f:
+        f.writelines(
+            [
+                "FRAPPE_VERSION=v14.17.1\n",
+                "ERPNEXT_VERSION=v14.9.0\n",
+                f"DB_PASSWORD={db_pass}\n",
+                "DB_HOST=db\n",
+                "DB_PORT=3306\n",
+                "REDIS_CACHE=redis-cache:6379\n",
+                "REDIS_QUEUE=redis-queue:6379\n",
+                "REDIS_SOCKETIO=redis-socketio:6379\n",
+                f"LETSENCRYPT_EMAIL={email}\n"
+                f"FRAPPE_SITE_NAME_HEADER={site_name}",
+            ]
+        )
+
+def generate_pass(length:int=9) -> str:
+    return b64encode(token_bytes(length)).decode()
+    
+
+def check_repo_exists() -> bool:
+    return os.path.exists(os.path.join(os.getcwd(), "frappe_docker"))
+
+
+def setup_prod(project: str, sitename: str,email:str):
+    if check_repo_exists():
+        compose_file_name = os.path.join(os.path.expanduser("~"), f"{project}-compose.yml")
+        docker_repo_path = os.path.join(os.getcwd(), "frappe_docker")
+        admin_pass = generate_pass()
+        db_pass = generate_pass(6)
+        print(
+            f"\n{CYLW}Please refer to .example.env file in the frappe_docker folder to know which keys to set{CEND}\n\n"
+        )
+        with open(compose_file_name, "w") as f:
+            if not os.path.exists(os.path.join(docker_repo_path, "/.env")):
+                write_to_env(docker_repo_path, sitename,db_pass,email)
+                print(
+                    f"\n{CYLW}A .env file is generated with basic configs. Please edit it to fit your needs {CEND}\n"
+                )
+            try:
+                # TODO: Include flags for non-https and non-erpnext installation
+                subprocess.run(
+                    [
+                        which("docker"),
+                        "compose",
+                        "--project-name",
+                        project,
+                        "-f",
+                        "compose.yaml",
+                        "-f",
+                        "overrides/compose.mariadb.yaml",
+                        "-f",
+                        "overrides/compose.redis.yaml",
+                        # "-f", "overrides/compose.noproxy.yaml", TODO: Add support for local proxying without HTTPs
+                        "-f",
+                        "overrides/compose.erpnext.yaml",
+                        "-f",
+                        "overrides/compose.https.yaml",
+                        "--env-file",
+                        ".env",
+                        "config",
+                    ],
+                    cwd=docker_repo_path,
+                    stdout=f,
+                    check=True,
+                )
+
+            except Exception as e:
+                print(f"\n{CRED}Generating Compose File failed{CEND}\n")
+        try:
+            subprocess.run(
+                [
+                    which("docker"),
+                    "compose",
+                    "-p",
+                    project,
+                    "-f",
+                    compose_file_name,
+                    "up",
+                    "-d",
+                ],
+                check=True,
+            )
+            
+        except Exception as e:
+            print(
+                f"{CRED} Docker Compose failed, please check the container logs{CEND}\n",
+                e,
+            )
+        print(f"\n{CGRN}Creating site: {sitename} {CEND}\n")
+        with open(os.path.join(os.path.expanduser("~"),"passwords.txt"),"w") as f:
+            f.writelines(f"Administrator:{admin_pass}\n")
+            f.writelines(f"MariaDB Root Password:{db_pass}\n")
+        try:
+            subprocess.run([
+                which("docker"),"compose",
+                "-p",project,"exec","backend",
+                "bench","new-site",sitename,
+                "--db-root-password",db_pass,
+                "--admin-password",admin_pass,
+                "--install-app","erpnext"
+                ],check=True)
+        except Exception as e:
+            print(
+                f"{CRED} Bench Site creation failed{CEND}\n",
+                e,
+            )
+    else:
+        install_docker()
+        clone_frappe_docker_repo()
+        setup_prod(project, sitename,email)  # Recursive
+
+
+def setup_dev_instance(project: str):
+    if check_repo_exists():
+        try:
+            subprocess.run(
+                [
+                    "docker",
+                    "compose",
+                    "-f",
+                    "devcontainer-example/docker-compose.yml",
+                    "--project-name",
+                    project,
+                    "up",
+                    "-d",
+                ],
+                cwd=os.path.join(os.getcwd(), "frappe_docker"),
+                check=True,
+            )
+            print(
+                f"{CYLW}Please go through the Development Documentation: https://github.com/frappe/frappe_docker/tree/main/development to fully complete the setup.{CEND}"
+            )
+        except Exception as e:
+            print(f"{CRED}Setting Up Development Environment Failed\n{CEND}", e)
+    else:
+        install_docker()
+        clone_frappe_docker_repo()
+        setup_dev_instance(project)  # Recursion on goes brrrr
+
+
+def install_docker():
+    if which("docker") is None:
+        print(f"{CGRN}Docker is not installed, Installing Docker...{CEND}")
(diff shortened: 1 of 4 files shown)


easy-install.py
```diff
@@ -0,0 +1,239 @@
...
+import time
+from shutil import which
+import platform
+from secrets import token_bytes
+from base64 import b64encode
+
+CRED = "\033[31m"
+CEND = "\033[0m"
+CGRN = "\33[92m"
+CYLW = "\33[93m"
```
ankush: Add function wrappers around these so we dont forget to add CEND :)

easy-install.py
```diff
@@ -0,0 +1,239 @@
...
+
+
+def clone_frappe_docker_repo() -> None:
+    try:
+        subprocess.run(
+            ["git", "clone", "https://github.com/frappe/frappe_docker"], check=True
```
ankush: This assumes git is installed, which by default isn't on most distros.

easy-install.py
```diff
@@ -0,0 +1,239 @@
...
+        print(f"\n{CRED}Cloning frappe_docker Failed{CEND}\n\n", e)
+
+
+def write_to_env(wd, site,db_pass,email):
+    site_name = site if site != "" else ""
+    with open(os.path.join(wd, ".env"), "w") as f:
+        f.writelines(
+            [
+                "FRAPPE_VERSION=v14.17.1\n",
+                "ERPNEXT_VERSION=v14.9.0\n",
```
ankush: ummm hardcoding versions?

ankush: Instead of hardcoding tags we can hard-code branches which don't change frequently. 

There's `v14` tag on docker hub which points to latest images in v14.x.x

https://hub.docker.com/layers/frappe/erpnext-worker/v14/images/sha256-500eb8a09edb45a698b42c1b42eb42441a7e8c1e44ac08d4c0746a72cd4526e5?context=explore

easy-install.py
```diff
@@ -0,0 +1,239 @@
...
+                "REDIS_CACHE=redis-cache:6379\n",
+                "REDIS_QUEUE=redis-queue:6379\n",
+                "REDIS_SOCKETIO=redis-socketio:6379\n",
+                f"LETSENCRYPT_EMAIL={email}\n"
+                f"FRAPPE_SITE_NAME_HEADER={site_name}",
+            ]
+        )
+
+def generate_pass(length:int=9) -> str:
+    return b64encode(token_bytes(length)).decode()
```
ankush: Note: `token_bytes` will give 2x length for specified length. If not intentional copy `generate_hash` from frappe to get exact length.

easy-install.py
ankush: This should be either part of instruction OR handled by script.

Alternatively, download the zip over HTTP and expand. We don't really need a git repository.

easy-install.py
```diff
@@ -0,0 +1,239 @@
...
+CYLW = "\33[93m"
+
+
+def clone_frappe_docker_repo() -> None:
+    try:
+        subprocess.run(
+            ["git", "clone", "https://github.com/frappe/frappe_docker"], check=True
+        )
+    except Exception as e:
+        print(f"\n{CRED}Cloning frappe_docker Failed{CEND}\n\n", e)
```
ankush: Log entire exception somewhere so we can ask for more info when required. Printing exception only prints final message. 


This applies everywhere in the script where exceptions are caught and generic message is printed. We'd need some input from those failures to fix them :)

easy-install.py
```diff
@@ -0,0 +1,239 @@
...
+
+
+def write_to_env(wd, site,db_pass,email):
+    site_name = site if site != "" else ""
```
ankush:
```suggestion
    site_name = site or ""
```

easy-install.py
```diff
@@ -0,0 +1,239 @@
...
+        help="Add email for the SSL.",
+        required="--prod" in sys.argv
+    )
+    args = parser.parse_args()
+    if args.dev:
+        print(f"\n{CGRN}Setting Up Development Instance{CEND}\n")
+        setup_dev_instance(args.project)
+    elif args.prod:
+        print(f"\n{CGRN}Setting Up Production Instance{CEND}\n")
+        if "example.com" in args.email:
```
ankush, line 320: why this specific check? Many dummy emails out there.

easy-install.py
```diff
@@ -0,0 +1,239 @@
...
+                Please go through the Docker Installation docs for your system and run this script again{CEND}"""
+            )
+            exit(1)
+        try:
+            ps = subprocess.run(
+                ["curl", "-fsSL", "https://get.docker.com"],
+                capture_output=True,
+                check=True,
+            )
+            subprocess.run(["/bin/bash"], input=ps.stdout, capture_output=True)
```
ankush: Run the first process as shell directly instead of manual redirection to shell?

easy-install.py
```diff
@@ -0,0 +1,239 @@
...
+            time.sleep(10)
+            subprocess.run(
+                ["sudo", "systemctl", "restart", "docker.service"], check=True
+            )
+        except Exception as e:
+            print(f"{CRED}Failed to Install Docker{CYLW}\n", e)
+            print(
+                f"\n\n {CYLW}Try Installing Docker Manually and re-run this script again{CEND}\n\n"
+            )
+            exit(1)
```
ankush: Use `sys.exit` 

`exit` global is for interactive scripts.

easy-install.py
```diff
@@ -0,0 +1,239 @@
...
+    else:
+        return
```
ankush:
```suggestion
```

This doesn't do anything

easy-install.py
```diff
@@ -0,0 +1,239 @@
...
+        except Exception as e:
+            print(f"{CRED}Setting Up Development Environment Failed\n{CEND}", e)
+    else:
+        install_docker()
+        clone_frappe_docker_repo()
+        setup_dev_instance(project)  # Recursion on goes brrrr
+
+
+def install_docker():
+    if which("docker") is None:
```
ankush: Use guard clause to avoid over-indentation. `if which("docker") is not None: return`


https://refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html

ankush (changes requested):

ankush:
TODO:
- [ ] Resolve open issues
- [ ] Manual tests @ankush 
- [ ] Add "copy and run in shell" 2-3 liners in README.md of bench and frappe
- [ ] Add non-erpnext install?

easy-install.py
ankush: use `logging` module with `exc_info=True`

Log as much as you can so we get decent input for fixing issues when they inevitably come.

ankush:
From "broad goal" POV we should also limit ourselves to few Linux distros. Ubutnu/Debian dervatives at first and maybe Fedora someday later. We can assume that other crazy distro users can setup frappe_docker themselves.

Easy install became a mess cause it was trying to handle too many things poorly.

easy-install.py
revant: These are already part of frappe_docker/example.env, and we have a release action bumping `example.env` on every frappe or erpnext release. Just copying `example.env` to `.env` can get the latest version without hardcoding.

read/update `.env` file after you copy from `example.env`

easy-install.py
revant: traefik will throw error if it is example.com for letsencrypt/certbot notification email.

easy-install.py
revant: > From "broad goal" POV we should also limit ourselves to few Linux distros. Ubutnu/Debian dervatives at first and maybe Fedora someday later. We can assume that other crazy distro users can setup frappe_docker themselves.

Once we have docker installed, distro doesn't matter.
If docker is not installed, only ubuntu/debian/fedora/centos/rhel or distros supported by `curl -fsSL get.docker.com | bash` will install docker. Any unsupported distro by the script shows message like:

```
ERROR: Unsupported distribution 'alpine'
```

more about script here https://github.com/docker/docker-install

In case of unsupported distro, if someone manages to install `docker` and `docker compose` then the script will proceed.

revant:
> From "broad goal" POV we should also limit ourselves to few Linux distros. Ubutnu/Debian dervatives at first and maybe Fedora someday later. We can assume that other crazy distro users can setup frappe_docker themselves.

Once we have docker installed, distro doesn't matter.
If docker is not installed, only ubuntu/debian/fedora/centos/rhel or distros supported by `curl -fsSL get.docker.com | bash` will install docker. Any unsupported distro by the script shows message like:

```
ERROR: Unsupported distribution 'alpine'
```

more about script here https://github.com/docker/docker-install

In case of unsupported distro, if someone manages to install `docker` and `docker compose` then the script will proceed.

easy-install.py
athul: If the version is provided as without minor version numbers, the latest image will be fetched. Say if I provide `FRAPPE_VERSION=v14`, the fetched image will be v14.17.1. Which is actually the latest version of Frappe rn(04/12/22)

easy-install.py
athul: Tried, didn't work 😓

athul:
Worked on all the suggestions and reviews, Tested out the script multiple times on a docker container inside a container(dind). You can too

```sh
# Download the easy install script to a folder called scripts
$ docker run --rm -it -v <path-to-scripts-folder>:/root/scripts -v /var/run/docker.sock:/var/run/docker.sock python:3-alpine sh
# Once inside the container
$ apk add docker-cli docker-cli-compose curl
$ cd ~/scripts
$ python easy-install.py --prod --project easyinstall -s easy.local --email <some-email>
```

revant:
I tested the script.

It installs with `v14` tag, That may pull latest v14 images causing unintended non-major upgrades. I'd recommend either get the version from frappe/erpnext releases or from example.env.

revant:
Everything works as expected. Site is up and running on `https://easy.localhost`, http redirects to https.
easy-install.log, easyinstall-compose.yml passwords.txt and frappe_docker directory found in path from where `easy-install.py` was executed.

add cronjob if possible https://github.com/frappe/frappe_docker/blob/main/docs/backup-and-push-cronjob.md

easy-install.py
```diff
@@ -0,0 +1,306 @@
...
+                "REDIS_SOCKETIO=redis-socketio:6379\n",
+                f"LETSENCRYPT_EMAIL={email}\n",
+                f"FRAPPE_SITE_NAME_HEADER={site_name}\n",
+                f"SITE_ADMIN_PASS={admin_pass}",
+            ]
+        )
+
+
+def generate_pass(length: int = 12) -> str:
+    return sha224(repr(time.time()).encode()).hexdigest()[:length]
```
ankush: time.time() isn't random enough. Copy the new version from develop branch :)

ankush:
```python
def generate_pass(length: int = 12) -> str:
	"""Generate random hash using best available randomness source."""
	import math
	import secrets

	if not length:
		length = 56

	return secrets.token_hex(math.ceil(length / 2))[:length]
```

revant:
Idea for cron yml template.

```yaml
# backup-job.yml.tmpl
services:
  backup:
    image: frappe/frappe-worker:${FRAPPE_VERSION:?No Frappe version set}
    volumes:
      - sites:/home/frappe/frappe-bench/sites
      - assets:/home/frappe/frappe-bench/sites/assets:ro
    entrypoint: ["bash", "-c"]
    command: |
      for SITE in $(/home/frappe/frappe-bench/env/bin/python -c "import frappe;print(' '.join(frappe.utils.get_sites()))")
      do
        bench --site $SITE backup --with-files
      done
    volumes:
      - sites:/home/frappe/frappe-bench/sites
    networks:
      - bench-network

networks:
  bench-network:
    external: true
    name: ${project}_default

volumes:
  sites:
    external: true
    name: ${project}_sites
```

`volumes` and `networks` need to be dynamic. How about making it as a template string python code? or read from template file from frappe_docker somewhere?

For upgrade?

- frappe_docker/.env needs to update the env vars or it'll just pick up existing ones from .env
- populate assets, the acrobatics of clearing volume and populating it so that python code can read js static files. Needed for `website_theme.py`
- migrate sites, `bench --site {site-name} migrate` to be executed for site(s) available in the frappe-bench?

more about populate assets and migrate sites here https://github.com/frappe/frappe_docker/issues/1020#issuecomment-1352626691

athul:
@revant Will add the cron in a subsequent PR. Thanks for the inputs 💟

ankush (approved):
Works flawlessly!

Awesome work @athul and thanks a lot for everything you do in Frappe ecosystem @revant 


:rocket:

PR #1399, opened 2022-12-09
fix: Add `--no-restart` flag to `init` and `get-app` bench commands

nk521:
Reason:
In FC, whenever you build a new deploy candidate, it used to fail with the latest bench version because of bench restarting services at the end of FF/any app installation (`supervisorctl restart frappe`). This flag forcefully stops that behavior.

After this change everything seems to work in FC.

### DIFF
--- a/bench/app.py
+++ b/bench/app.py
@@ -346,6 +346,7 @@ def get_app(
 	soft_link=False,
 	init_bench=False,
 	resolve_deps=False,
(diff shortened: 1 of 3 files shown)


nk521:
Need to run some test at FC end. After that I'll mark this ready.

ankush:
handled by https://github.com/frappe/bench/pull/1400

PR #1402, opened 2022-12-12
ci: fix broken tests

ankush:
ci: run tests against current hotfix branches


ankush:
We are spending 9-10 minutes just on `bench build` on v13 tests :woozy_face: 

IMO we can trust framework works and skip assets in bench tests (?)

Issue #1338, opened 2022-07-28
Introduce `bench log` to deal with all-things-log


surajshetty3416:
We should have `bench log` commands to pull and display logs available in bench.

- It should show last n line of specific type of commands
```bash
bench log --source worker -n 100
bench log --source scheduler -n 100
bench log --source web.error -n 100
```

- It should also show logs related to specific sites
```bash
bench --site sitename log --source web -n 100
```

- It should have the ability to stream (or tail) log files
```bash
bench --site sitename log --source web -n 100 --stream
```

- It can have filters (grep, from, to, etc.)
```bash
bench --site sitename log --source web.error -n 100 --stream --find "ValidationError:"
bench --site sitename log --source web --from "2022-02-02" --to "2022-05-02"
```

PR #1412, opened 2023-01-11
Add `log` command

athul:
What type of a PR is this?

- [x] New Feature Submissions
- [ ] Bug Fix

---

This is a response to #1338 . Hopefully this feature will help folks but since this is in early stages. 

- Implemented the `log` command. 
Users can use `bench log frappe -n 30` to check the logs of `frappe.log` file inside the logs folder in bench. Implemented using a subprocess call to tail command

Screenshot 2023-01-11 at 3 19 27 PM

- Implemented `log -ls` for listing out all the log files in the log folder.

Screenshot 2023-01-11 at 3 20 46 PM

---

To add
- Streaming the log file(maybe use the -f flag of tail?)
- Output to another
- Docs

### DIFF
--- a/bench/commands/__init__.py
+++ b/bench/commands/__init__.py
@@ -89,6 +89,7 @@ from bench.commands.utils import (
 	set_ssl_certificate_key,
 	set_url_root,
 	start,
+	print_logs
 )
 
 bench_command.add_command(start)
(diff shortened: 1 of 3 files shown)


gavindsouza:
Site filter would be nice too

also if this command adds anything that simply using tail doesn't already do - it should be treating all paginated files as one?

ankush:
@gavindsouza 

`cat frappe.log*`

Whatever you add there's probably a superior unix alternative. It's just hard to beat it for text exploration. :woozy_face:

Issue #1414, opened 2023-01-12
apps.txt isn't maintaining order like before. 


ankush:
`apps.txt` roughly maintained `get-app *` order. Right now it seems to be random. 

https://github.com/frappe/bench/blob/34197056ea7de6735ea0fd8fc63b18093b3f634b/bench/bench.py#L274-L282

ankush:
Fixed in framework

PR #1415, opened 2023-01-13
fix: apps.txt should maintain order

ankush:
Problem: bench is storing apps.txt in random order, this is because every time `bench.apps.sync` is called it does `os.listdir()` to re-initialize apps.txt which gives output in practically random order.

Solution: Enforce invariants everywhere.

1. Initialize apps from disk by reading `apps.txt`
2. Any operation that modifies `apps.txt` should update `bench.apps` e.g. `install-app` does `bench.apps.append(...)`, `remove-app` does `bench.apps.remove(...)`.
3. Once in-memory state is changed, `bench.apps.sync()` should "flush" the changes to apps.txt file. This way it doesn't need to re-initialize the apps from disk

Potential caveats to check/address:
- [ ] If operation fails in between sync and install/remove apps.txt will not be in-sync.
- [ ] Dependency resolution code probably isn't maintaining this invariant?
- [ ] CLI is initialized after `Bench` is created, so non-bench directories are problamatic.

### DIFF
--- a/bench/bench.py
+++ b/bench/bench.py
@@ -17,7 +17,6 @@ from bench.utils import (
 	paths_in_bench,
 	exec_cmd,
 	is_bench_directory,
-	is_frappe_app,
 	get_cmd_output,
 	get_git_version,
 	log,
@@ -62,12 +61,12 @@ class Bench(Base, Validator):
 		self.cwd = os.path.abspath(path)
 		self.exists = is_bench_directory(self.name)
 
+		self.apps_txt = os.path.join(self.name, "sites", "apps.txt")
+		self.excluded_apps_txt = os.path.join(self.name, "sites", "excluded_apps.txt")
+
 		self.setup = BenchSetup(self)
 		self.teardown = BenchTearDown(self)
 		self.apps = BenchApps(self)
-
-		self.apps_txt = os.path.join(self.name, "sites", "apps.txt")
-		self.excluded_apps_txt = os.path.join(self.name, "sites", "excluded_apps.txt")
 
 	@property
 	def python(self) -> str:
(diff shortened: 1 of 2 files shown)


bench/bench.py
```diff
@@ -62,13 +61,13 @@ def __init__(self, path):
 		self.cwd = os.path.abspath(path)
 		self.exists = is_bench_directory(self.name)
 
+		self.apps_txt = os.path.join(self.name, "sites", "apps.txt")
+		self.excluded_apps_txt = os.path.join(self.name, "sites", "excluded_apps.txt")
```
ankush, line 65: Moved from 3 line below to here because:
1. these two steps don't depend on any of them
2. `self.apps` creation now depends on `self.apps_txt`

bench/bench.py
```diff
@@ -262,7 +260,6 @@ def sync(
 	):
 		if required == UNSET_ARG:
 			required = []
-		self.initialize_apps()
```
ankush, line 265: This was the source of randomness in ordering. `self.apps` is already present and "updated" with whatever changes are made to it.

revant:
I'm doing this in case of (re)configuring apps.txt

```
ls -1 apps > sites/apps.txt 
```

if at all it's merged, also make it into a bench command

```
bench config apps
# or anything suitable
```

ankush:
@revant it's largely not required. Only a few things like translations, command resolution and asset building depend on it. 

https://github.com/frappe/frappe/pull/19653

ankush:
laters

casesolved-co-uk:
Why was this abandoned?
Required for `bench update` / `bench setup requirements`

#1414 
https://discuss.frappe.io/t/how-to-change-sequence-of-applying-multiple-custom-apps/50462/6

Of course the real answer is app dependency resolution order, but maintaining manual ordering is a step in the right direction

PR #1419, opened 2023-01-18
fix: git app ssh username may not be git

casesolved-co-uk:
Bug fix for ssh git app server usernames that do not start with "git@"

backport:
v5.x

### DIFF
--- a/bench/app.py
+++ b/bench/app.py
@@ -129,7 +129,7 @@ class AppMeta:
 
 	def __setup_details_from_git(self, url=None):
 		name = url if url else self.name
-		if name.startswith("git@") or name.startswith("ssh://"):
+		if "@" in name or name.startswith("ssh://"):
 			self.use_ssh = True
 			_first_part, _second_part = name.rsplit(":", 1)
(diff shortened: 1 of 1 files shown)


bench/app.py
```diff
@@ -129,7 +129,7 @@ def _setup_details_from_git_url(self, url=None):
 
 	def __setup_details_from_git(self, url=None):
 		name = url if url else self.name
-		if name.startswith("git@") or name.startswith("ssh://"):
+		if "@" in name or name.startswith("ssh://"):
```
ankush, line 132: This breaks clones with PAT which do have `@` in them and follow this format:

`git clone https://{USER}:{PAT}@github.com/{REPO}`


I'd suggest using `ssh://` if you need to use non-`git` user.

ankush (changes requested):

bench/app.py
ankush: Alternatively reorder conditions so HTTP clones have higher priority and both cases get handled.

PR #1426, opened 2023-02-05
doc: add nix installation

blaggacao:
can be merged as soon as is merged: https://github.com/NixOS/nixpkgs/pull/214800

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,7 @@ Bench is a command-line utility that helps you to install, update, and manage mu
 		- [Arguments](#arguments)
 		- [Troubleshooting](#troubleshooting)
 	- [Manual Installation](#manual-installation)
+	- [Installation from Nixpkgs(#installation-from-nixpkgs)
 - [Basic Usage](#basic-usage)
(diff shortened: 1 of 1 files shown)


blaggacao:
By default, `nixpkgs` errors out on [unmaintained dependencies](https://github.com/frappe/bench/issues/1427). So I had to remove it. Users can still override and opt-in into a looming security risk.

ankush:
Let's keep it out of this repo. 

I appreciate the effort but these things require long-term maintenance too, so even if I merge it to repo there's no way to guarantee it will work even 2 months from today. 


IMO we should just stick to `frappe_docker` based installs for production use.

blaggacao:
@ankush Absolutely understood! There is a real possibility that a nix can help reduce your overall maintenance load on `bench`, but obviously, that would need a significant up-front climb on the learning curve, so consider this a first point of contact / context :smile:

I have done some work in a private repo in that regard, not sure yet when ready to share, and also not sure when it's ready: how to share. Maybe best in discourse, first...

Anyway thanks for letting me know where you're heads are at w.r.t. this.

PR #1439, opened 2023-02-25
fix: add custom background workers in group

SaiFi0102:
**Problem**
`bench restart` does not restart **custom background worker queues** because they're not part of the `{{ bench_name }}-workers` supervisor group

**Solution**
add custom background worker queues in the worker groups as well


SaiFi0102:
@ankush

ankush (approved):

PR #1446, opened 2023-04-21
fix: Handle apps with different repo & app names

gavindsouza:
### Changes

* Handle `remove-app` when Name and Repo names don't match
* Update relevant usages to use app_name over repo attr of `App`

### DIFF
--- a/bench/app.py
+++ b/bench/app.py
@@ -268,7 +268,7 @@ class App(AppMeta):
 			except IndexError:
 				return []
 		try:
-			required_deps = get_required_deps(self.org, self.app_name, self.tag or self.branch)
+			required_deps = get_required_deps(self.org, self.repo, self.tag or self.branch)
 			return required_apps_from_hooks(required_deps)
 		except Exception:
 			return []
--- a/bench/app.py
+++ b/bench/app.py
@@ -198,7 +198,7 @@ class App(AppMeta):
 
 	@step(title="Archiving App {repo}", success="App {repo} Archived")
 	def remove(self, no_backup: bool = False):
-		active_app_path = os.path.join("apps", self.repo)
+		active_app_path = os.path.join("apps", self.app_name)
 
 		if no_backup:
 			if not os.path.islink(active_app_path):
@@ -209,7 +209,7 @@ class App(AppMeta):
 		else:
 			archived_path = os.path.join("archived", "apps")
 			archived_name = get_available_folder_name(
-				f"{self.repo}-{date.today()}", archived_path
+				f"{self.app_name}-{date.today()}", archived_path
 			)
 			archived_app_path = os.path.join(archived_path, archived_name)
 
--- a/bench/app.py
+++ b/bench/app.py
@@ -233,7 +233,7 @@ class App(AppMeta):
 
 		verbose = bench.cli.verbose or verbose
 		app_name = get_app_name(self.bench.name, self.app_name)
-		if not resolved and self.repo != "frappe" and not ignore_resolution:
+		if not resolved and self.app_name != "frappe" and not ignore_resolution:
 			click.secho(
 				f"Ignoring dependencies of {self.name}. To install dependencies use --resolve-deps",
 				fg="yellow",
@@ -262,13 +262,13 @@ class App(AppMeta):
 		from bench.utils.app import get_required_deps, required_apps_from_hooks
 
 		if self.on_disk:
-			required_deps = os.path.join(self.mount_path, self.repo, "hooks.py")
+			required_deps = os.path.join(self.mount_path, self.app_name, "hooks.py")
 			try:
 				return required_apps_from_hooks(required_deps, local=True)
 			except IndexError:
 				return []
 		try:
-			required_deps = get_required_deps(self.org, self.repo, self.tag or self.branch)
+			required_deps = get_required_deps(self.org, self.app_name, self.tag or self.branch)
 			return required_apps_from_hooks(required_deps)
 		except Exception:
 			return []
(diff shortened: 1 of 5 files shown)


bench/cli.py
```diff
@@ -240,3 +242,17 @@ def _chdir(*args, **kwargs):
...
+
+	def handle_exception(exc_type, exc_info, tb):
+		print("".join(generate_exc(exc_type, exc_info, tb)))
+
+	def generate_exc(exc_type, exc_info, tb):
+		for t in format_exception(exc_type, exc_info, tb):
+			if "/click/" not in t:
+				yield t
+
+	sys.excepthook = handle_exception
```
phot0n: interesting: this replaces the `apport_excepthook` in ubuntu/ubuntu based distros (like popos)

https://wiki.ubuntu.com/Apport
https://wiki.ubuntu.com/ErrorTracker

before the change:
Screenshot from 2023-05-07 18-50-16

#

stumbled upon this by chance when i had not applied this change and wanted to check if the execepthook will spill over to a frappe process invoked from bench  (which it doesn;t btw {maybe due to venv? I have yet to try it out in a simple script} - was curious as the docs mentioned the [environment is copied](https://docs.python.org/3/library/os.html#os.execv) {brain instantly went to fork-like thing})

bench/cli.py
```diff
@@ -240,3 +242,17 @@ def _chdir(*args, **kwargs):
...
+def setup_exception_handler():
+	from traceback import format_exception
+
+	def handle_exception(exc_type, exc_info, tb):
+		print("".join(generate_exc(exc_type, exc_info, tb)))
+
+	def generate_exc(exc_type, exc_info, tb):
+		for t in format_exception(exc_type, exc_info, tb):
+			if "/click/" not in t:
+				yield t
```
phot0n: could some breakage in click go unnoticed due to this?

bench/bench.py
```diff
@@ -131,6 +131,9 @@ def uninstall(self, app, no_backup=False, force=False):
 		except InvalidRemoteException:
 			if not force:
 				raise
+		except ValueError:
+			pass
```
phot0n: escaped as the `repo` name doesnt exist in self.apps

#

~should we also remove it from `self.apps` over here?~

https://github.com/frappe/bench/blob/f3b55f7785710cf293aaa00a087b1bd7e760e7cf/bench/bench.py#L318

currently it's fine because, the `.sync()` call below re-inits the apps sequence before flushing to apps.txt

phot0n (commented):
`app_name` is needed over here as well?

https://github.com/frappe/bench/blob/f3b55f7785710cf293aaa00a087b1bd7e760e7cf/bench/app.py#L263

bench/cli.py
gavindsouza: I've been using Pop for over a year and haven't encountered the `apport_excepthook`. Nothing like you've shared pops up in my terminal lol

It doesn't spill over to Frappe/other app commands since the bench process (interpreter 1 - global) replaces itself with a new invocation (interpreter 2 - env) by the `EXECV` directive with "copied" `os.environ`.

bench/cli.py
gavindsouza: Wondered this too. Perhaps this change should only come into effect for certain Bench Exceptions? Would appreciate thoughts from more users on this.

Should only click be excluded or are there more packages that add to the noise, which commands are more susceptible to this?

phot0n: let's revert this for now? - bunch of additional frames in traceback don't really hurt that much as opposed to if some breakages goes unnoticed? :eyes:

gavindsouza: The goal was to get rid of the noise which leads to confusion for newbies and non-tech users. One part of this was to print the failing command rather than a bare exception. The other is to not show an overwhelming amount of context in case of failure.

click is pretty robust - I've not seen breakages in ~5 years of using it. So I am more confident about excluding it from the traceback. I'd rather not see a CLI showing tracebacks like this at all. So, developing in that direction, I think we can apply this traceback filtering conditionally - exclude frames from internals in case "X" exception was raised...in this case - `CommandFailedError`

phot0n: also, this would ignore tracebacks from a folder/package named "click" :P

I'm personally fine with the change per say (because i dont have any folders named click and mostly dont run into bench errors that often *now*) but the fact that it has the potential to "hide" so much stuff is a bit daunting to me. I'd rather have more info shown (be it unnecessary) than less info.

bench/bench.py
gavindsouza: Changed it anyway (for correctness)

bench/app.py
```diff
@@ -262,13 +262,13 @@ def _get_dependencies(self):
...
 		if self.on_disk:
-			required_deps = os.path.join(self.mount_path, self.repo, "hooks.py")
+			required_deps = os.path.join(self.mount_path, self.app_name, "hooks.py")
 			try:
 				return required_apps_from_hooks(required_deps, local=True)
 			except IndexError:
 				return []
 		try:
-			required_deps = get_required_deps(self.org, self.repo, self.tag or self.branch)
+			required_deps = get_required_deps(self.org, self.app_name, self.tag or self.branch)
```
phot0n: this should be `.repo`? we're querying github using it

https://github.com/frappe/bench/blob/3e448d4d42ed25caa8da2547a151a6db89b617b4/bench/utils/app.py#L183

bench/bench.py
phot0n: this can be removed now? or is this kept for people who've already installed apps with different repo names?

gavindsouza:
Moved changes related to CLI cleanup into a separate PR - now this branch contains only App.app_name/App.repo related fixes

bench/app.py
gavindsouza: Oops

bench/bench.py
gavindsouza: I'm actually unsure, how about handling it the same way as InvalidRemoteException above `(raise if not force)`?

gavindsouza: Removed via d9c8335

PR #1450, opened 2023-05-03
fix: resolve filesystem app dependencies

### LINKED ISSUE #1449: Bench fails to resolve dependencies when using file system path to install app
Issue: Bug report

**Do the checklist before filing an issue:**
 - [x] Can you replicate the issue on the supported bench versions?
 - [x] Is this something you can debug and fix? Send a pull request! Bug fixes and documentation fixes are welcome

**Describe the bug** :chart_with_downwards_trend:

I have created a frappe custom_app
- It's private repo or a repo I haven't pushed to GitHub. Basically, not accessible by GitHub API publically.
- I want to install the custom_app locally using file-system path.

When installing the custom_app using the command `bench get-app --resolve-deps /home/frappe/custom_app`. 

Bench doesn't detects the dependencies required by the custom_app. It only installs the custom_app.

But when installing `hrms` (which has public git repo) using the command `bench get-app --resolve-deps hrms`. It successfully detects the required dependencies and installs them.

**To Reproduce** :page_with_curl:
1. Create a new app using `bench new-app custom_app`
2. Mention the dependencies in `apps/custom_app/custom_app/hooks.py` file. eg. ERPNext
3. Copy the custom_app  to any other directory. eg. `/home/custom_app`
4. Remove this app from bench using `bench rm custom_app`.
5. Install custom_app using `bench get-app --resolve-deps /home/custom_app`

Here, you will notice that the custom_app will get installed but not ERPNext.

**Expected behavior** :chart_with_upwards_trend:
It should read the `hooks.py` file from the filesystem instead of trying to fetch it from GitHub API.

**OS (please complete the following information):** :cyclone:
- [x] Linux: `Ubuntu:22.04.2 LTS`

**Version Information**
```shell
$ bench --version
5.16.2
```

**Possible Solution** :bookmark_tabs:
`_get_dependencies` function work is to get the required dependencies of the the app and returns it as list type.

if `self.on_disk` is true then only the dependencies are read from the `hooks.py` file of custom_app.

But during the execution flow of installing custom_app, `self.on_disk` remains `false`.

So as the execution continues, `get_required_deps`  tries to fetch `hooks.py` using GitHub API. As, the app is not public, it will not be able to fetch `hooks.py`. Resulting in `404: NOT Found` that leads to not installing the dependencies (this issue).

Xieyt:
fixes #1449 
Please check GitHub issue for details

[GitHub Actions report on forked repo](https://github.com/Xieyt/bench/actions)

- [ ] Docs have been added / updated (not required)
- [ ] Tests for the changes have been added (for bug fixes / features) (not required)
- [ ] Did you modify the existing test cases? If yes, why? (not required)

### DIFF
--- a/bench/app.py
+++ b/bench/app.py
@@ -81,16 +81,18 @@ class AppMeta:
 		if not self.is_repo:
 			self.repo = self.app_name = self.name
 			return
+
+		# fetch meta for repo on mounted disk
+		if os.path.exists(self.mount_path):
+			self.on_disk = True
+			self._setup_details_from_mounted_disk()
+
 		# fetch meta from installed apps
-		if self.bench and os.path.exists(os.path.join(self.bench.name, "apps", self.name)):
+		elif self.bench and os.path.exists(os.path.join(self.bench.name, "apps", self.name)):
 			self.mount_path = os.path.join(self.bench.name, "apps", self.name)
 			self.from_apps = True
 			self._setup_details_from_mounted_disk()
 
-		# fetch meta for repo on mounted disk
-		elif os.path.exists(self.mount_path):
-			self.on_disk = True
-			self._setup_details_from_mounted_disk()
 
 		# fetch meta for repo from remote git server - traditional get-app url
 		elif is_git_url(self.name):
--- a/bench/app.py
+++ b/bench/app.py
@@ -82,15 +82,11 @@ class AppMeta:
 			self.repo = self.app_name = self.name
 			return
 
-		# fetch meta for repo on mounted disk
-		if os.path.exists(self.mount_path):
-			self.on_disk = True
-			self._setup_details_from_mounted_disk()
-
 		# fetch meta from installed apps
-		elif self.bench and os.path.exists(os.path.join(self.bench.name, "apps", self.name)):
+		if self.bench and os.path.exists(os.path.join(self.bench.name, "apps", self.name)):
 			self.mount_path = os.path.join(self.bench.name, "apps", self.name)
 			self.from_apps = True
+			self.on_disk = True
 			self._setup_details_from_mounted_disk()
 
 


bench/app.py
```diff
@@ -81,16 +81,18 @@ def setup_details(self):
...
+
 		# fetch meta from installed apps
-		if self.bench and os.path.exists(os.path.join(self.bench.name, "apps", self.name)):
+		elif self.bench and os.path.exists(os.path.join(self.bench.name, "apps", self.name)):
 			self.mount_path = os.path.join(self.bench.name, "apps", self.name)
 			self.from_apps = True
 			self._setup_details_from_mounted_disk()
 
-		# fetch meta for repo on mounted disk
-		elif os.path.exists(self.mount_path):
```
ankush, line 91: How does changing the order of conditions fix this? 

Is first condition also true previously? :thinking: 

If so, can't we just set self.on_disk to true in that condition?

Xieyt: >Is first condition also true previously?

Yes. The [previous first condition](https://github.com/frappe/bench/blob/develop/bench/app.py#L85-L88) was getting true previously.

>If so, can't we just set self.on_disk to true in that condition?

I tested it and this will work as well. In that case I think [this condition](https://github.com/frappe/bench/blob/develop/bench/app.py#L91-L93) won't be necessary.

---

I have found another bug while testing the above. - If dependent apps are already installed, new app installation fails while running the `get-app` command with `--resolve-deps`.
Eg. If I already have ERPNext installed in the Bench, and I try installing custom_app then the installation fails. Here, custom_app has ERPNext as dependency. Same goes for installing `hrms` with `--resolve-deps`.

I have fixed this new bug as well, testing currently.

ankush (approved):

amalpaul54111:
Looks like this PR included some breaking changes for the bench , This could be verified by checking at the failed CI/Tests https://github.com/frappe/bench/actions/runs/4881678149/jobs/8712432796. This is causing errors when running setup-requirements. This has also been raised as an separate issue in https://github.com/frappe/bench/issues/1451.

PR #1481, opened 2023-09-05
fix: sync socketio port for old benches

ankush:
closes https://github.com/frappe/bench/pull/1382#issuecomment-1706495743


SaiFi0102:
Thank you Ankush!

PR #1489, opened 2023-10-04
refactor(treewide): use `mariadb` commands and service instead of `mysql`

### LINKED ISSUE #1472: mysql: Deprecated program name. It will be removed in a future release, use '/usr/bin/mariadb' instead
When working with MariaDB, this software uses the executable `mysql` which is deprecated. The following error message occurs.

```
mysql: Deprecated program name. It will be removed in a future release, use '/usr/bin/mariadb' instead
```

You need to switch to `mariadb`.

akhilnarang:
refactor: use `mariadb` commands and service instead of `mysql`

This resolves #1472

This has been done because the mysql command issues a deprecation warning now
mysql: Deprecated program name. It will be removed in a future release, use '/usr/bin/mariadb' instead

### DIFF
--- a/bench/commands/setup.py
+++ b/bench/commands/setup.py
@@ -358,7 +358,7 @@ def sync_domains(domain=None, site=None):
 @click.command("role", help="Install dependencies via ansible roles")
 @click.argument("role")
 @click.option("--admin_emails", default="")
-@click.option("--mysql_root_password")
+@click.option("--mysql_root_password", "--mariadb_root_password")
 @click.option("--container", is_flag=True, default=False)
 def setup_roles(role, **kwargs):
 	extra_vars = {"production": True}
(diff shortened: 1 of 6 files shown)


bench/commands/setup.py
```diff
@@ -358,7 +358,7 @@ def sync_domains(domain=None, site=None):
 @click.command("role", help="Install dependencies via ansible roles")
 @click.argument("role")
 @click.option("--admin_emails", default="")
-@click.option("--mysql_root_password")
+@click.option("--mariadb_root_password")
```
ankush: We shouldn't change this, this can be breaking change for some scripts. 

Add another alias for same flag? IIRC `click` does support that.

ankush (changes requested):

bench/commands/setup.py
akhilnarang: Ah cool, will do that.
Wanted to ask anyway, there's a lot of reference to MySQL in general in a few places like this, so for now should we leave them as-is or add secondary option with mariadb like this?

akhilnarang: Added as aliases. Left variable names and dict keys intact so that it won't cause any issues.

ankush (approved):

PR #1519, opened 2024-01-15
feat: cache get-app artifacts by commit_hash

### LINKED ISSUE #1242: perf(build): Better cache pip and yarn packages
I am not sure if this already works, on surface it seems like it doesn't.

yarn packages should be cached _outside_ so next build if it misses app install cache entirely will just use locally available and more importantly **built** packages. 

This is just insane:

image

Possible fix: https://stackoverflow.com/questions/65913706/how-do-i-make-yarn-cache-modules-when-building-containers https://github.com/moby/buildkit

18alantom:
TLDR: Stores `get-app` artifacts in `.cache` by ~commit hash~ cache key and reuses them when applicable.

Partially Fixes: https://github.com/frappe/press/issues/1242

**TODO:**
- [x] Cache output and restore cache if present.
- [x] Add explicit `compress_artifacts` flag.
- [x] Update `frappe/esbuild/esbuild.js` to account for cached get-app. https://github.com/frappe/frappe/pull/24412
- [x] Complete installation from cache.
- [x] Ensure `$BENCH/sites` is being updated properly.
- [x] Get rid of `node_modules` for apps with built frontends (gameplan, insights, etc).
- [x] Commands to clear, and view app cache.
- [x] Check if it works.

### Usage

Adds two new options to `bench get-app`:
- `--cache-key`: Used to store tarred app folder in `~/.cache/bench/apps` with the following format `"{app}-{cache-key[:10]}.tar"`
- `--compress-artifacts`: Optional flag that causes artifacts to be compressed using gzip, extension will then be `.tgz`.

**Examples:**

```bash
# Get from local repo, apps/gampelan will be cached if not present
bench get-app file:///home/frappe/gameplan --cache-key 87d4aa3b10aa23a620d5a51a968ae47110c43dd1

# Get from remote frappe repo, apps/erpnext will be compressed and cached if not present
bench get-app erpnext --cache-key 7bcea6099da1a41ad530bdd85ff337f627039b80 --compress-artifacts
```

For either of the above if app is present in cache it will be fetched from there, else it will be stored in cache. While installing, cache is checked for both compressed and non compressed tar files.

### Utility Commands

<details>
<summary><code>bench app-cache</code></summary>

Screenshot 2024-01-19 at 19 10 18

</details>

<details>
<summary><code>bench app-cache --remove-app APP_NAME</code></summary>

Screenshot 2024-01-19 at 19 10 40

</details>

<details>
<summary><code>bench app-cache --remove-key CACHE_KEY</code></summary>

Screenshot 2024-01-19 at 19 11 23

</details>

### Why?

When running `get-app` bench spends significant amount of time in git clone, yarn install, pip install, building frontend dependencies. If it's running get-app on the same code (same app, same repo, same commit hash), `get-app` artifacts from the previous run can be reused.

### How?

After `get-app` has completed for some `$APP` the relevant folders that change are:
- `$BENCH/apps/$APP`: cloned source files, and built frontend files and dependencies.
- `$BENCH/env`: python dependencies for the get-app'd `$APP`.
- `$BENCH/sites`: link to the `assets/$APP` folder, update `assets.json`, `apps.json`, `apps.txt`.

Caching the contents of `$BENCH/apps/$APP` covers repo cloning, yarn install and yarn build. That's what this PR facilitates.

<details>
<summary>Reference</summary>

File system changes after running `get-app` for `frappe/hrms`.

Screenshot 2024-01-15 at 11 49 18

</details>

**Regarding `$BENCH/env`**

Pip install does take a considerable amount of time but since it isn't contained in the app folder it won't be included for now. Future updates might have python dependencies stored inside the app folder. Out of scope for now.

### DIFF
--- a/bench/app.py
+++ b/bench/app.py
@@ -6,10 +6,12 @@ import re
 import shutil
 import subprocess
 import sys
+import tarfile
 import typing
 from collections import OrderedDict
 from datetime import date
 from functools import lru_cache
+from pathlib import Path
 from urllib.parse import urlparse
 
 # imports - third party imports
@@ -19,16 +21,11 @@ import git
 # imports - module imports
 import bench
 from bench.exceptions import NotInBenchDirectoryError
-from bench.utils import (
-	UNSET_ARG,
-	fetch_details_from_tag,
-	get_available_folder_name,
-	is_bench_directory,
-	is_git_url,
-	is_valid_frappe_branch,
-	log,
-	run_frappe_cmd,
-)
+from bench.utils import (UNSET_ARG, fetch_details_from_tag,
+                         get_available_folder_name, get_bench_cache_path,
+                         is_bench_directory, is_git_url,
+                         is_valid_frappe_branch, log, run_frappe_cmd)
+from bench.utils.app import check_existing_dir
 from bench.utils.bench import build_assets, install_python_dev_dependencies
 from bench.utils.render import step
 
@@ -166,6 +163,7 @@ class App(AppMeta):
 		branch: str = None,
 		bench: "Bench" = None,
 		soft_link: bool = False,
+		commit_hash = None,
 		*args,
 		**kwargs,
 	):
@@ -173,6 +171,7 @@ class App(AppMeta):
 		self.soft_link = soft_link
 		self.required_by = None
 		self.local_resolution = []
+		self.commit_hash =  commit_hash
 		super().__init__(name, branch, *args, **kwargs)
 
 	@step(title="Fetching App {repo}", success="App {repo} Fetched")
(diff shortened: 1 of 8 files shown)


ankush:
>  Get rid of node_modules for apps with built frontends (gameplan, insights, etc).

This will not allow rebuild btw (we allow doing it from FC dashboard :grimacing: )

bench/app.py
```diff
@@ -283,6 +290,90 @@ def update_app_state(self):
...
+		cwd = os.getcwd()
+		cache_path = self.get_app_cache_path(compress_artifacts)
+		mode =  "w:gz" if compress_artifacts else "w"
+		
+		message = "Caching get-app artifacts"
+		if compress_artifacts:
+			message += " (compressed)"
+		click.secho(message)
+
+		os.chdir(app_path.parent)
```
ankush, line 370: If possible avoid changing directories. This has caused problems in past as bench assumes a certain directory to be cwd always.

18alantom:
> > Get rid of node_modules for apps with built frontends (gameplan, insights, etc).
> 
> This will not allow rebuild btw (we allow doing it from FC dashboard 😬 )

Ah, I didn't know we allowed rebuilds. I suppose we can trigger `yarn install` if `node_modules` not found. It will increase the first rebuild time, but seems like a reasonable tradeoff. What do you think?

bench/app.py
18alantom: Here it's required cause of how tar functions, it stores the entire path and untarring requires moving the file out of nested directories. I've wrapped tar in a `try` `finally` so `cwd` should be restored irrespective: https://github.com/frappe/bench/pull/1519/commits/30f301e3ffa8a2561160ec740d2d046355905c97

18alantom:
**Note: ignoring SonarCloud Quality Gate**

It's failing due to potential tar bombs. It's failed in two places when creating the archive and when extracting it.

**When creating an archive**: check is not necessary here cause the archive file is newly created and will be empty. Else that stack of code would not have run.

**When extracting from an archive**: archive being extracted from is created by bench, and not user submitted, well the app repo could be, but in the given context it seems highly unlikely.

(that being said I might add an `extractall` filter in a subsequent PR, it is not going to appease SonarCloud but will address the issue)


_Edit: filter added in https://github.com/frappe/bench/pull/1523_

Issue #1524, opened 2024-01-31
feat: Frappe package dependencies


ankush:
We can specify python dependencies in `pyproject.toml`

```toml
dependencies = [
    # Core dependencies
    "pycountry~=22.3.5",
    "Unidecode~=1.3.6",
    "barcodenumber~=0.5.0",
    "rapidfuzz~=2.15.0",
    "holidays~=0.28",
]
```

However we can't use Frappe app versions here because they are not published on pypi and pip can possibly fuckup and try to fetch it from pypi and fail catastrophically. (ref https://github.com/frappe/erpnext/pull/25501)

Possible solution:

1. We use same syntax as requirements.txt https://pip.pypa.io/en/stable/reference/requirements-file-format/
2. We add a special section in pyproject.toml

```toml
[tool.bench.frappe-dependencies]
frappe = ">=15.2.0"
erpnext = "~=15.1.1"
```

3. When doing `bench get-app` bench ensures that dependencies are present before installing app.  

Note: This won't resolve and install them at all. (Not ideal for platforms like FC where we can't just install arbitrary app dependencies) That can be handled some other day.

Possible problems: 
- Frappe is installed first but other apps aren't really providing any guarantee for ordering. Order of installation can matter here. No idea how to solve this. Mostly, in Frappe ecosystem currently Frappe and ERPNext version targeting is required first.

Example usage of custom pyproject config blocks:
https://github.com/frappe/bench/blob/d84a05ed2a9e21ff7efd8bff98b14d13d8bacc18/bench/utils/bench.py#L110

PR #1525, opened 2024-01-31
fix: delete node_modules if frappe supports app_cache

18alantom:
Since part of the build executes in frappe, get-app cache requires at least v15.12.0 of frappe to function optimally i.e. delete node_modules and skip build. This PR checks frappe version mentioned in the app's `pyproject.toml`.

Uses version specifier format mentioned in: https://github.com/frappe/bench/issues/1524

### DIFF
--- a/bench/app.py
+++ b/bench/app.py
@@ -12,6 +12,7 @@ from collections import OrderedDict
 from datetime import date
 from functools import lru_cache
 from pathlib import Path
+from typing import Optional
 from urllib.parse import urlparse
 
 # imports - third party imports
@@ -170,7 +171,7 @@ class App(AppMeta):
 		branch: str = None,
 		bench: "Bench" = None,
 		soft_link: bool = False,
-		cache_key = None,
+		cache_key=None,
 		*args,
 		**kwargs,
 	):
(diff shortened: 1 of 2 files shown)


ankush:
@18alantom just few steps away from closing linked issue, might as well complete it :grimacing:

18alantom:
> @18alantom just few steps away from closing linked issue, might as well complete it 😬

@ankush I'm unsure about what all needs to be done w.r.t the linked issue:
- How should dependency absence be handled, fail get-app or just show a warning?
- Assuming `frappe-dependencies` are runtime, i.e. get-app can still run without the `frappe-dependencies`, doesn't make sense to fail it.
- Any way to indicate strict requirement during get-app, (in case `setup.py` needs the dependency).

If we're looking to fail get-apps if mentioned dependency is not present, we should hold off until more thought is given. Press Deploy Candidate app ordering too would have to be updated to account for it.

ankush:
Yes we should fail with non-zero exit code.

If ordering is a hard problem to solve we can optionally add a check command to do it after all apps are installed on bench?

18alantom:
Added dependency check, fails if folder not present. Also added a version check, doesn't fail but gives a warning if mismatch.

Regarding ordering:
1. Ordering needs to be managed by whomever is installing apps using bench.
2. If (all) apps have frappe-dependencies mentioned, invalid ordering is not possible (cause it will fail).
3. If apps do not have frappe-dependencies mentioned, invalid ordering cannot be prevented (else get-app breaks).
4. By point 2. and 3. ordering check in bench is redundant.

PR #1533, opened 2024-02-07
fix: Check yarn files during install

ankush:
We observed that in some cases files are missing when yarn reuses them
from cache. Not sure what's causing it but checking files seems to at
least fix it. This is probably some yarn bug.

https://github.com/yarnpkg/yarn/issues/3223#issuecomment-296373273


vinaymavi:
This check-files option is not supported by latest yarn relase and depreciated in 1.x version.

PR #1537, opened 2024-02-21
fix(build): pin hatchling version to prevent import error

18alantom:
reference: https://github.com/pypa/hatch/pull/1255

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,7 +44,7 @@ Source = "https://github.com/frappe/bench"
 
 [build-system]
 requires = [
-    "hatchling>=1.6.0",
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -44,7 +44,7 @@ Source = "https://github.com/frappe/bench"
 
 [build-system]
 requires = [
-    "hatchling>=1.6.0",
+    "hatching>=1.6.0,<1.9.x",
```
ankush, line 47: L

ankush:
```suggestion
    "hatchling>=1.6.0,<1.9.4",
```

PR #1538, opened 2024-02-21
fix(build): pin hatchling version 2nd try

18alantom:
back ref: https://github.com/frappe/bench/pull/1537

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,7 +44,7 @@ Source = "https://github.com/frappe/bench"
 
 [build-system]
 requires = [
-    "hatchling>=1.6.0,<1.9.4",
+    "hatchling>=1.6.0,<1.7.0",
 ]
 build-backend = "hatchling.build"
 


18alantom:
This too doesn't seem to have worked, the issue is perhaps elsewhere.

ofek:
What is the issue?

18alantom:
Hey @ofek our [release script](https://github.com/frappe/bench/actions/runs/7988847272/job/21814274565#step:6:206) was failing with:

```
ImportError: cannot import name 'EDITABLES_REQUIREMENT' from 'hatchling.builders.constants'
```

due it installing `hatch` from the master branch. This does not work cause the changes in `hatch/pull/1255` causes the breakage (due to lack of familiarity with python build systems, I'm not entirely sure why).

As a workaround, pinning `hatch` to 1.9.2 seems to have worked.

(The previous tries did not work because I was blindly changing the `hatchling` version)

PR #1555, opened 2024-05-06
fix: check cache hash before using

18alantom:
Check hash md5 against previously calculated md5 (when setting cache), improve a few of the bench app-cache messages.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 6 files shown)


18alantom:
Ignoring SonarCloud cause it's complaining about MD5 not being safe and to use SHA512 instead.
This misses the context as MD5 is being used here for integrity check not security.

ankush:
> Ignoring SonarCloud cause it's complaining about MD5 not being safe and to use SHA512 instead.

Add `usedforsecurity=False` https://docs.python.org/3/library/hashlib.html


(maybe not, 3.9+ only :woozy_face: )

PR #1556, opened 2024-05-08
fix: add useforsecurity with fallback for <3.9

18alantom:
this change has been added, primarily, to trigger a release


18alantom:
https://github.com/frappe/bench/pull/1555

PR #1583, opened 2024-09-18
feat: mTLS with nginx and python

HarryPaulo:
When we are going to integrate 2 systems where we need a high level of security between the two ends of the communication, we must use mTLS to verify and validate that the requests received from the client have the certificate...
Normally this type of integration does not work with HMAC, it must use mTLS.

image

So, I looked for the most efficient and flexible configuration to apply this, thinking that on a server with a "bench" we will have more than one site and each site can have more than one integration enabled with mTLS certificate configured and each integration can have your own certificate

image

The ngnix configuration was chosen:
ssl_verify_client **optional_no_ca**
**optional_no_ca**: requests the client certificate but does not require it to be signed by a trusted CA certificate. This is intended for the use in cases when a service that is external to nginx performs the actual certificate verification. The contents of the certificate is accessible through the [$ssl_client_cert](https://nginx.org/en/docs/http/ngx_http_ssl_module.html#var_ssl_client_cert) variable.

"proxy_set_header SSL_CLIENT_CERT $ssl_client_cert;": allows you to get the certificate from the header, using a relative variable.
image

So with this config, i can made a validade on Python loading the certificate configured on relative integration and confirming the identity:

image

ps: Maybe there are some more points to be implemented, I will be putting my application into production in a few days or weeks...

I am open to suggestions

### DIFF
--- a/bench/config/templates/nginx.conf
+++ b/bench/config/templates/nginx.conf
@@ -48,6 +48,7 @@ server {
 	ssl_ciphers EECDH+AESGCM:EDH+AESGCM;
 	ssl_ecdh_curve secp384r1;
 	ssl_prefer_server_ciphers on;
+	ssl_verify_client optional_no_ca;
 	{% endif %}
 
 	add_header X-Frame-Options "SAMEORIGIN";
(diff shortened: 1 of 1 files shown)


HarryPaulo:
I'm working on it, and I think this first change wasn't a good one because it's in the base url location / {...}

I'm studying and exploring something that fits better.

PR #1587, opened 2024-10-14
feat: allow nginx apply mtls for a specific path

HarryPaulo:
For this change to work on the bench, it is necessary to analyze the frappe pull request:
https://github.com/frappe/frappe/pull/28125

Currently, some integrations that work with sensitive data are required to use mTLS, for example integration with credit card companies, banks, fintechs...

mTLS consists of validating whether the client is using a valid certificate to access the service, but it is a two-way validation, in this case it validates the client and server certificates.

image

Requests to the URL "/api/secure/{methodname}" will require mandatory validation at both ends of the digital certificate, thus applying mTLS validation by nginx.

**ssl_verify_client optional** was chosen because this method does not impact other URLs.

### DIFF
--- a/bench/commands/config.py
+++ b/bench/commands/config.py
@@ -55,6 +55,14 @@ def config_http_timeout(seconds):
 	update_config({"http_timeout": seconds})
 
 
+@click.command(
+	"ssl_client_certificate", help="Set certificate path to nginx use on mtls"
+)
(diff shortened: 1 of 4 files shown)


HarryPaulo:
We solved this using mTLS with amazon API gateway... maybe this can be reopened on the future.

PR #1590, opened 2024-10-26
fix: use optional-dependencies as preferred variant to custom dev-dependencies

blaggacao:


blaggacao:
cc @gavindsouza see https://github.com/frappe/frappe/blob/5b718bb60f856229ea5caddf51955d513d7ed96a/pyproject.toml#L99-L135

`pip install -e .[dev,test]`

bench/utils/bench.py
```diff
@@ -97,8 +97,15 @@ def install_python_dev_dependencies(bench_path=".", apps=None, verbose=False):
...
-			pyproject_deps = _generate_dev_deps_pattern(pyproject_path)
-			if pyproject_deps:
+			try:
+				from tomli import loads
+			except ImportError:
+				from tomllib import loads
+		
+			pyroject_config = loads(open(pyproject_path).read())
+			for key, deps in pyroject_config["project"]["optional-dependencies"].items():
+				bench.run(f"{bench.python} -m pip install {quiet_flag} --upgrade {' '.join(deps)}")
```
gavindsouza, line 107: I suppose this could just be moved into the function that installs the apps instead

`bench.run(f"{bench.python} -m pip install {quiet_flag} --upgrade -e './apps/{app}[{', '.join(pyroject_config["project"]["optional-dependencies"])}]'`

Let pip worry about the entire env management

Given we are using the standard to define optional dependencies, I wonder if we could just retire `bench.dev-dependencies` in favour of the standard.

blaggacao: > I suppose this could just be moved into the function that installs the apps instead

If "install app" can be equated to "install optional deps, as well", then yes, indeed.

> retire `tools.bench.dev-dependencies`

Yes that makes sense, absolutely!

gavindsouza: For development setups, installing optional dependencies should be ok, although if I go through the effort of making separate groups, I would expect more control. 

So, I'm not sure automatically installing the different "optional-dependencies" groups would be the best course of action. Passing `--install-optional-dependencies=dev,test` sounds like a saner path to take, also adding a warning/info if developer_mode is enabled on bench and optional-dependencies are present in app pyproject.toml.

kevinli02-02 (approved):

PR #1600, opened 2024-11-26
feat: add more options to profcile render

blaggacao:
This PR increases the flexibility to render procfile according to needs.

### DIFF
--- a/bench/config/procfile.py
+++ b/bench/config/procfile.py
@@ -4,12 +4,14 @@ import platform
 import click
 
 import bench
-from bench.app import use_rq
 from bench.bench import Bench
 from bench.utils import which
 
 
-def setup_procfile(bench_path, yes=False, skip_redis=False):
+def setup_procfile(bench_path, yes=False, skip_redis=False, skip_web=False, skip_watch=None, skip_socketio=False, skip_schedule=False, with_coverage=False):
+	if skip_watch is None:
+		# backwards compatibilty; may be eventually removed
+		skip_watch = os.environ.get("CI")
 	config = Bench(bench_path).conf
 	procfile_path = os.path.join(bench_path, "Procfile")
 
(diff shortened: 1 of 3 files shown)


bench/config/procfile.py
```diff
@@ -4,12 +4,11 @@
 import click
 
 import bench
-from bench.app import use_rq
 from bench.bench import Bench
 from bench.utils import which
 
 
-def setup_procfile(bench_path, yes=False, skip_redis=False):
+def setup_procfile(bench_path, yes=False, skip_redis=False, skip_web=False, skip_watch=os.environ.get("CI"), skip_socketio=False, skip_schedule=False, with_coverage=False):
```
akhilnarang: Can we set `None` for skip_watch and then set os.environ.get() inside? Have had a weird experience setting default values like this (might not affect this specific case because its a CLI that would be freshly imported/called everytime, but may as well follow that pattern everywhere)

akhilnarang (changes requested):

bench/config/procfile.py
blaggacao: https://github.com/frappe/bench/compare/95182c89ed62720628dd97e1e1a4002f9bbdcec2..786b63def57e582d508917a6ca5d13a11be1a8b3

bench/config/templates/Procfile
```diff
@@ -2,15 +2,18 @@
 redis_cache: redis-server config/redis_cache.conf
 redis_queue: redis-server config/redis_queue.conf
 {% endif %}
-web: bench serve {% if webserver_port -%} --port {{ webserver_port }} {%- endif %}
-
+{% if not skip_web %}
+web: bench serve {% if with_coverage -%} --with-coverage  {%- endif %} {% if webserver_port -%} --port {{ webserver_port }} {%- endif %}
```
akhilnarang, line 6: Any cases where webserver_port was null here?

blaggacao: This was in there before, but I've not personally observed a case where this would run on `80`

akhilnarang: Ah, I misread the diff, somehow I didn't see that it was there earlier 🤣

akhilnarang (changes requested):
All good functionality wise, but should move to use the existing click commands that we have

akhilnarang (approved):

PR #1602, opened 2024-12-07
fix: --no-mariadb-socket is DEPRECATED

0xD0M1M0:
--no-mariadb-socket is DEPRECATED; use --mariadb-user-host-login-scope='%' (wildcard) or --mariadb-user-host-login-scope=, instead.

### DIFF
--- a/easy-install.py
+++ b/easy-install.py
@@ -416,7 +416,7 @@ def create_site(
         "backend",
         "bench",
         "new-site",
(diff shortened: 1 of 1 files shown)


akhilnarang:
Its fine to leave it in for now, that flag won't be dropped anytime soon for backwards compatibility reasons. Can't switch it here either for similar reasons - the "new" flag isn't available in older versions of framework.

PR #1609, opened 2024-12-25
feat(easy-install): backup cron

revant:
- [x] Docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,8 +97,8 @@ options:
   -h, --help            show this help message and exit
   -n PROJECT, --project PROJECT
                         Project Name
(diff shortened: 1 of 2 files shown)


NagariaHussain:
Nice!

The `upgrade` command could be simplified with defaults for image and version:

```sh
./easy-install.py upgrade --project=insights_prod_setup --image=ghcr.io/frappe/insights --version=stable --backup-schedule="@every 2m"
```

Also can we have a better way to update frappe_docker contents itself? Deleting and re-creating (script does that) resulted in loss of .env file. Or should we move this file somewhere else?

Issue #1610, opened 2024-12-26
easy-install upgrade command


NagariaHussain:
The `upgrade` command could be simplified with defaults for image and version:

```sh
./easy-install.py upgrade --project=insights_prod_setup --image=ghcr.io/frappe/insights --version=stable --backup-schedule="@every 2m"
```

Also can we have a better way to update frappe_docker contents itself? Deleting and re-creating (script does that) resulted in loss of .env file. Or should we move this file somewhere else?

_Originally posted by @NagariaHussain in https://github.com/frappe/bench/issues/1609#issuecomment-2562240823_

revant:
- [x] move image and tag from in-memory variables to `.env` file
- [x] move frappe_docker/.env to `~/<project-name>.env`
- [x] move `~/passwords.txt` to `~/<project-name>-passwords.txt`

PR #1611, opened 2024-12-26
fix(easy-install): read values from env if exists

revant:
#1610

### DIFF
--- a/easy-install.py
+++ b/easy-install.py
@@ -172,8 +172,10 @@ def start_prod(
     db_pass = ""
     custom_image = None
     custom_tag = None
+
     if image:
-        [custom_image, custom_tag] = image.split(":")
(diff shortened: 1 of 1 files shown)


NagariaHussain:
Will test today 👍

revant:
No change in docs. Command remains same, explained here: https://github.com/frappe/bench?tab=readme-ov-file#arguments

NagariaHussain:
Cool, will test and merge today 👍

NagariaHussain:
I tried running the `upgrade` command with an updated backup cron string. It did update the env, but not the actual container it seems:

CleanShot 2024-12-31 at 12 34 12@2x

CleanShot 2024-12-31 at 12 34 02@2x

PR #1622, opened 2025-03-18
feat(bench): miscellaneous improvements for pkg-config

akhilnarang:
- ensure `pkg-config` is installed
- Set `PKG_CONFIG_PATH` for macOS

### DIFF
--- a/bench/app.py
+++ b/bench/app.py
@@ -909,13 +909,15 @@ def install_app(
 
 	app_path = os.path.realpath(os.path.join(bench_path, "apps", app))
 
-	command_prefix = ""
+	env = {}
 	# macOS needs a custom PKG_CONFIG_DIR for frappe
 	if app == "frappe" and sys.platform == "darwin":
-		command_prefix = "PKG_CONFIG_PATH=$(brew --prefix mariadb-connector-c)/lib/pkgconfig"
+		env = {
+			"PKG_CONFIG_PATH": "/opt/homebrew/opt/mariadb-connector-c/lib/pkgconfig",
+		}
 
 	bench.run(
-		f"{command_prefix} {bench.python} -m pip install {quiet_flag} --upgrade -e {app_path} {cache_flag}"
+		f"{bench.python} -m pip install {quiet_flag} --upgrade -e {app_path} {cache_flag}", env=env
 	)
 
 	if conf.get("developer_mode"):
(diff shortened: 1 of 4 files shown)


bench/bench.py
```diff
@@ -415,6 +423,15 @@ def wheel(self, verbose=False):
 			f"{self.bench.python} -m pip install {quiet_flag} wheel", cwd=self.bench.name
 		)
 
+	@step(title="Checking if pkg-config is installed", success="pkg-config is present")
+	def check_pkg_config(self):
+		"""
+		pkg-config is required for building some python packages like libmysqlclient
+		"""
+		if shutil.which("pkg-config") is None:
+			raise Exception("pkg-config is not installed. Please install it before proceeding.\n"
```
gavindsouza: Maybe do this only for develop/v16+?

bench/app.py
```diff
@@ -909,8 +909,15 @@ def install_app(
 
 	app_path = os.path.realpath(os.path.join(bench_path, "apps", app))
 
+	env = {}
+	# macOS needs a custom PKG_CONFIG_DIR for frappe
+	if app == "frappe" and sys.platform == "darwin":
+		env = {
+			"PKG_CONFIG_PATH": "/opt/homebrew/opt/mariadb-connector-c/lib/pkgconfig",
```
gavindsouza: This looks brittle

Maybe something like `pkg-config --variable=pc_path mariadb-connector-c` or `$(brew --prefix mariadb-connector-c)/lib/pkgconfig` and ask user to set it manually if envvar is not set properly

akhilnarang: Hmm yeah my previous version used brew --prefix mariadb-connector-c to get it dynamically, had done this to quickly test something while I had access to a Mac, will re-add that part with this logic soon.

bench/bench.py
gavindsouza: Without this, this is a breaking change

akhilnarang: > Without this, this is a breaking change

Hmm fair. Will add in a check.

akhilnarang: We just pip install {app_name} for requirements setup, need to see how to get version in sanely.

akhilnarang:
Everything _should_ be fine, it is when I ssh in and manually run the commands, just pip install within the test suite can't find pkg-config somehow, which shouldn't be related to this PR.

Edit: `subprocess.call(env={})` was the cause

PR #1632, opened 2025-05-11
fix!: use check_nested_command, drop python 3.9

### LINKED ISSUE #1633: Strict Click Versions
We need to have a strict `Click` version since a major version update fails to resolve:

`from click.core import _check_multicommand_`

sokumon:
Click released a new version 8.2.0
It doesnt have the _check_multicommand function as it has been renamed
Ref https://github.com/pallets/click/blob/8.2.0/src/click/core.py

```pip install frappe-bench```  fails because of this 

> [!CAUTION]
> This PR drops support for python3.9.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ['3.9', '3.10', '3.11', '3.12', '3.13']
+        python-version: ['3.10', '3.11', '3.12', '3.13']
 
     name: Base (${{ matrix.python-version }})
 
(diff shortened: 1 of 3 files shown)


akhilnarang:
We should also pin the click version.

sokumon:
Should we pin it to 8.2.0 or the 8.1.8

revant (approved):

barredterra:
I guess `~=8.2.0` would be correct because that's where the renamed function exists.

barredterra:
Shall we follow click in dropping support for python3.9? I'd say yes.

barredterra:
I think the easy install check fails because the error that this PR tries to fix is part of the latest docker image that is used in the test case. So the failure should be safe to ignore.

akhilnarang:
> Shall we follow click in dropping support for python3.9? I'd say yes.

v14+ requires py310+, so seems fine

PR #1634, opened 2025-05-11
fix: Set stricter click version to ~=8.1.x

### LINKED ISSUE #1633: Strict Click Versions

maheini:
closes #1633 

Does currently break all custom frappe_docker builds because of the new release of Click 8.0!

- [ ] New Feature Submissions
- [x] Bug 

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,7 +19,7 @@ classifiers = [
     "Topic :: System :: Installation/Setup",
 ]
 dependencies = [
-    "Click~=7.0",
+    "Click~=8.1.0",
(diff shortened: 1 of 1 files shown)


barredterra:
While this PR fixes the issue, https://github.com/frappe/bench/pull/1632 contains a more comprehensive fix that also adds support for the newer Click version.

maheini:
Awesome, this will be even better in the long run :smiley: 

Thank you.

PR #1637, opened 2025-05-21
fix(bench): support python3.8/3.9 as well

akhilnarang:


akhilnarang:
v13 users can just use `frappe-bench==5.23.0`

PR #1657, opened 2025-09-07
feat: add bench show-ports command to list site ports

nirajregmi:
What type of PR is this?

 New Feature Submissions

 Changes to Existing Features

 Bug Fix

 Breaking Change (include change in API behaviours, etc.)

Details / Problem this PR solves

Currently, Frappe developers have no easy way to check which sites are running on which NGINX ports. Typically, they need to manually open config/nginx.conf and inspect the configuration. This is tedious, error-prone, and not convenient, especially for benches with multiple sites.

This PR introduces a new bench command:

bench show-ports

Features:

Lists all sites and their configured NGINX ports.

Supports filtering a single site:

bench show-ports site1.local

Gracefully handles missing nginx.conf.

Works with Python 3.10+ and Click.

Uses get_bench_name() correctly with the bench path.

Usage Examples

List all sites and ports in the bench:

$ bench show-ports
Bench my-bench sites and ports:
site1.local → Port 80
site2.local → Port 443

Filter by a single site:

$ bench show-ports --site site1.local
site1.local → Port 80

When nginx.conf is missing:

$ bench show-ports
No nginx.conf found. Try running `bench setup nginx` first.

Real-life Scenario

Imagine a developer managing a Frappe bench with 10 different sites. Some sites run on default port 80, others on 443, and maybe a few custom ports. During deployment, debugging, or testing, they need to quickly verify which site is mapped to which port — especially when multiple sites share the same domain with different ports.

Before this feature, they would have to open config/nginx.conf and manually search for each site, which is slow and error-prone. With bench show-ports, they can instantly see all site-port mappings in one command, making site management, debugging, and configuration much faster and reliable.

### DIFF
--- a/bench/commands/__init__.py
+++ b/bench/commands/__init__.py
@@ -92,6 +92,7 @@ from bench.commands.utils import (
 	set_ssl_certificate_key,
 	set_url_root,
 	start,
+    show_ports
 )
 
 bench_command.add_command(start)
@@ -112,6 +113,7 @@ bench_command.add_command(bench_src)
 bench_command.add_command(find_benches)
 bench_command.add_command(migrate_env)
 bench_command.add_command(app_cache_helper)
+bench_command.add_command(show_ports)
 
 from bench.commands.setup import setup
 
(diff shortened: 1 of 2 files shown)


akhilnarang:
Parsing nginx like this seems too error-prone.

From what I've seen, site config has an `nginx_port` which is then used to generate the config. Using that seems like a better approach.

PR #1661, opened 2025-09-13
fix: add frappe app warnings to bench start

YashReddy1963:
### What this PR does
This PR adds a warning to the `bench start` command for developers when installed Frappe apps are detected in non-editable mode (.egg or site-packages). 

- The logic is implemented in `frappe/utils/bench_helper.py` (Frappe app).
- The call is placed in `bench/commands/utils.py` inside the `start()` command. 

Please refer to issue #33385
Docs: No-Docs

### DIFF
--- a/bench/commands/utils.py
+++ b/bench/commands/utils.py
@@ -5,6 +5,7 @@ import os
 import click
 
 
+
 @click.command("start", help="Start Frappe development processes")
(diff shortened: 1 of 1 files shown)


akhilnarang:
Had mentioned `bench serve` instead of `bench start` as that's within frappe codebase and not in bench. Better not to have cross-dependencies like this.

PR #1662, opened 2025-10-06
feat: integrate mprocs

safwansamsudeen:
Adds two configuration options to `bench init`:
- `--use-mprocs`: allows people to use `mprocs` instead of of `honcho` - though both are still supported even with this option
- `--default-app TEXT`: allows developers to set a default app to run the frontend server for

`bench start` will now use `mprocs` if `mprocs.yaml` exists - otherwise it'll use honcho (current working).

Also, if `default-app` is specified, both `use_mprocs` and `install-app` are assumed so that this:
```
bench init --use-mprocs --install-app drive --default-app drive PATH
```
becomes:
```
bench init --default-app drive PATH
```

### DIFF
--- a/bench/bench.py
+++ b/bench/bench.py
@@ -399,7 +399,7 @@ class BenchSetup(Base):
 					)
 
 	@step(title="Setting Up Bench Config", success="Bench Config Set Up")
-	def config(self, redis=True, procfile=True, additional_config=None):
+	def config(self, redis=True, procfile=True, use_mprocs=False, additional_config=None, default_app=None):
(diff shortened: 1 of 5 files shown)


akhilnarang (commented):
While its fine overall, "default app" sounds weird. Maybe can tweak that flag name. 

Also I've noticed many of the apps READMEs had mentioned using `yarn`, here `npm` is hardcoded.

I'd prefer adding in some parameter which basically lets you specify a service name + command, and that would get added to procfile/mprocs yaml as chosen.

PR #1663, opened 2025-10-06
feat: Enhance Python executable discovery in bench.py

asaura08:
Enhancements to Python executable resolution:

* Added logic to check for the Python executable in `~/.pyenv/versions/<version>/bin/python` if the provided Python argument looks like a version number and is not found in the system path.
* Improved error handling to raise a clear exception if the Python executable cannot be found, guiding the user to check their PATH or provide a full path.

This pull request improves the robustness of the `migrate_env` function in `bench/utils/bench.py` by enhancing how the Python executable is located. Now, if the specified Python version is not found in the system path, the function attempts to locate it within the user's `pyenv` installations before raising an error.

Now you can run `bench migrate-env 3.13.7`

---

What type of a PR is this?

- [x] Changes to Existing Features
- [ ] New Feature Submissions
- [ ] Bug Fix

---

### DIFF
--- a/bench/utils/bench.py
+++ b/bench/utils/bench.py
@@ -203,7 +203,19 @@ def migrate_env(python, backup=False):
 	bench = Bench(".")
 	nvenv = "env"
 	path = os.getcwd()
+	original_python = python
 	python = which(python)
+	
+	if not python:
+		# Try finding python in pyenv if it's a version number
+		if original_python and re.match(r"^\d+\.\d+\.\d+$", original_python):
+			pyenv_path = os.path.expanduser("~/.pyenv/versions")
+			possible_path = os.path.join(pyenv_path, original_python, "bin", "python")
+			if os.path.exists(possible_path):
+				python = possible_path
+				
+	if not python:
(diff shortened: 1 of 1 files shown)


akhilnarang:
No point hardcoding this for one specific tool. If you're using a tool to manage python, it should have some command that would print the full path.

```bash
bench migrate-env `uv python find 3.13`
bench migrate-env `uv python find 3.14`
```

I regularly migrate like this.

asaura08:
> No point hardcoding this for one specific tool. If you're using a tool to manage python, it should have some command that would print the full path.
> 
> ```shell
> bench migrate-env `uv python find 3.13`
> bench migrate-env `uv python find 3.14`
> ```
> 
> I regularly migrate like this.

I do this because when use the dev container [frappe/bench:latest](https://hub.docker.com/layers/frappe/bench/latest/images/sha256-e13095e937ab96c3b5d7a385710478bc45731cd8db5834bdf63cd6bea3da897d) it come with pyenv tool, so maybe for developers it's useful

akhilnarang:
`pyenv which python` should do the job to get the path?

asaura08:
Yes, this command gets the path, I just wanted to code a more intuitive way

PR #1664, opened 2025-10-09
fix: app cache failure fallback

akhilnarang:
- fix: only change git_url if app cache can actually be extracted
- build(deps): bump


Aradhya-Tripathi:
LGTM

PR #1669, opened 2025-11-03
feat(app): App name from ast

Aradhya-Tripathi:
- Parse the `ast` to get the app-name
- Current regex pattern is too greedy

### DIFF
--- a/bench/utils/app.py
+++ b/bench/utils/app.py
@@ -1,20 +1,23 @@
 # imports - standard imports
+import ast
 import os
 import pathlib
 import re
-import sys
 import subprocess
-from typing import List, Optional
+import sys
+import warnings
 from functools import lru_cache
+from typing import List, Optional
+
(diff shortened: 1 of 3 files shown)


bench/utils/app.py
```diff
@@ -222,6 +227,18 @@ def get_remote(app, bench_path="."):
 		return contents.splitlines()[0].split()[0]
 
 
+def get_app_name_from_setup(contents: str) -> str:
```
akhilnarang:
```suggestion
def get_app_name_from_setup(contents: str) -> str | None:
```

Or `return ""` or raise some exception at the end after the loop/checks?

akhilnarang (changes requested):

bench/utils/app.py
Aradhya-Tripathi: #1669 0a75989

akhilnarang (approved):

PR #1678, opened 2025-12-19
security(nginx): remove OCSP stapling configuration

ismxilxrif:
Let’s Encrypt has ended support for OCSP responses and OCSP stapling. As a result, ssl_stapling and ssl_stapling_verify no longer provide any security benefit and may cause warnings or misleading configuration state in modern Nginx deployments.

This change removes OCSP stapling directives from the nginx template to align with current TLS practices and browser revocation mechanisms (Certificate Transparency / CRLite).

Impact:
- No reduction in TLS security
- Avoids dead or misleading configuration
- Prevents potential startup or reload warnings
- Aligns with Let’s Encrypt post-OCSP revocation model

Reference:
https://letsencrypt.org/2024/12/05/ending-ocsp

 ---

image

### DIFF
--- a/bench/config/templates/nginx.conf
+++ b/bench/config/templates/nginx.conf
@@ -42,8 +42,6 @@ server {
 	ssl_session_timeout  5m;
 	ssl_session_cache shared:SSL:10m;
 	ssl_session_tickets off;
-	ssl_stapling on;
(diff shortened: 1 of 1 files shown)


ismxilxrif:
can be backported to branch v5.x

PR #1697, opened 2026-02-06
fix(easy-install): pull tagged frappe_docker instead of main

DanielRadlAMR:
What type of a PR is this?

- [ ] New Feature Submissions
- [x] Bug Fix

---

`frappe_docker` recently introduced breaking changes. This PR is a hotifx, it pulls the latest tag before the breaking change, instead of the main branch. This is not a final solution, as the easy-install script should be migrated correctly.

#1696

### DIFF
--- a/easy-install.py
+++ b/easy-install.py
@@ -46,13 +46,13 @@ def cprint(*args, level: int = 1):
 def clone_frappe_docker_repo() -> None:
     try:
         urllib.request.urlretrieve(
-            "https://github.com/frappe/frappe_docker/archive/refs/heads/main.zip",
+            "https://github.com/frappe/frappe_docker/archive/refs/tags/v1.0.1.zip",
             "frappe_docker.zip",
         )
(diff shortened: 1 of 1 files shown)


asieftejani:
Hi @DanielRadlAMR
I am checking the changes - I think there is a mistake - it should be v1.0.0 (which has image: traefik:v2.11 in https://github.com/frappe/frappe_docker/blob/v1.0.0/overrides/compose.proxy.yaml) and not v1.0.1(which has image: traefik:v3.6 in https://github.com/frappe/frappe_docker/blob/v1.0.1/overrides/compose.proxy.yaml)
Please can you double chek and confim

DanielRadlAMR:
`v1.0.1` was the lates version **before** the breaking change.

Although it runs on Traefik `v3`, the appropriate compatibility flags are enabled to preserve `v2` syntax behavior.

DanielRadlAMR:
#1724

PR #1700, opened 2026-02-10
chore(easy-install): update defaults to match frappe_docker

DanielRadlAMR:
Switched `easy-install` defaults to frappe `v16` with corresponding Python and Node version


DanielRadlAMR:
#1724

PR #1716, opened 2026-06-12
feat(socketio): Runtime backend selection via `bench socketio`

tanmoysrt:
Related : https://github.com/frappe/frappe/pull/39879


rmehta:
Did you fix this in frappe/bench-cli as well?

tanmoysrt:
> Did you fix this in frappe/bench-cli as well?

I have a draft PR on bench-cli. https://github.com/frappe/bench-cli/pull/43 for the same.
But, I need to make few more changes in that PR for actual benefit of memory usage from the changes : https://github.com/frappe/gunicorn/pull/15

rmehta:
cool, maybe just focus on the way forward with the new bench, we might not take this to v2 anyway.

PR #1718, opened 2026-06-13
feat(config): Add gunicorn companion process manager

tanmoysrt:
Add dedicated gunicorn.py config module and template alongside updates to supervisor and systemd configs to support gunicorn as an alternative web process manager alongside the existing setup.


tanmoysrt:
will be added in new bench-cli
https://github.com/frappe/bench-cli

PR #1721, opened 2026-06-16
fix: Don't update everything while installing application

### LINKED ISSUE #1683: uv pip upgrade behavior breaks multi-app install workflow and violates existing dependency constraints
Bench’s current app installation workflow assumes that dependencies are resolved incrementally and that versions installed in earlier steps remain compatible when later apps are installed. This was true when Bench relied on `pip`, but it no longer holds after the switch to `uv`.

**The Problem**
Bench installs apps one by one through separate commands

With `pip`, this typically preserved earlier installed versions unless a conflicting requirement appeared. With `uv`, the resolver treats only the active requirement set, not already installed distributions, so shared dependencies may be upgraded beyond constraints required by previously installed apps.

**Example**

In version-16-beta:

`frappe` requires: `oauthlib ~=3.2.2` (allowed: `>=3.2.2, <3.3.0`)

`erpnext` (via `python-youtube` → `requests-oauthlib`) requires: `oauthlib >=3.0.0`

Installing both in one resolution step picks a compatible version (`3.2.x`).
Installing in two separate steps causes uv pip to upgrade `oauthlib` to `3.3.2`, violating Frappe’s requirement, because the constraint `~=3.2.2` is not part of the visible input during the second resolution pass.

**Why this is a regression**

The existing workflow relies on sequential installation steps.

The switch to `uv` breaks this model because constraints from already installed apps are not imported into subsequent resolution runs.

Although it's not causing problems now, this might break apps in the future.

This produces environments that Bench previously avoided without explicit intervention.

**Impact**

Multi-app benches can end up with incompatible shared dependencies after adding apps.

**Expected behavior**

Bench should maintain dependency consistency across all installed apps without requiring manual  management from users.

ankush:
fixes https://github.com/frappe/bench/issues/1683

### DIFF
--- a/bench/app.py
+++ b/bench/app.py
@@ -934,8 +934,9 @@ def install_app(
 
 	if use_uv():
 		try:
+			# Avoid upgrade flag  on a fresh install; see #1683
 			bench.run(
-				f"uv pip install {quiet_flag} --upgrade -e {app_path} {cache_flag} --python {bench.python}",
+				f"uv pip install {quiet_flag} -e {app_path} {cache_flag} --python {bench.python}",
 				env=env,
 			)
 		except Exception as e:
--- a/bench/bench.py
+++ b/bench/bench.py
@@ -34,7 +34,7 @@ from bench.utils.bench import (
 	get_env_cmd,
 )
 from bench.utils.render import job, step
-from bench.utils.app import get_current_version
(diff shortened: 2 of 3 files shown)


ankush:
Unsure of all the possible side effects.

@copilot should I merge?

PR #1723, opened 2026-06-22
feat(completions): add filesystem path tab completion

barredterra:
Goal: `bench restore [TAB]` and similar commands should offer file path completions. 

Collects path-taking options and positional arguments from Click metadata and completes them with `compgen -f`, covering commands like `restore`, `backup`, and `data-import` without breaking existing site/app completion.

Changes CI to actually run completion tests and install zsh to make them pass.

Contains some noise from formatting python files (`style:` commit) and refactoring complex functions to satisfy Sonarqube's complexitiy score (`refactor: ... Cognitive Complexity`).


barredterra:
Easy install test failure is tackled in #1724

PR #1701, opened 2026-02-19
fix(easy-install): support Traefik v3 SITES_RULE and add safe site mismatch handling

Rocket-Quack:
#1696 
Fixed because frappe-docker now accepts a different format for site_rules due to traefik updates.

## Summary
This PR updates `easy-install.py` for Traefik v3-compatible site routing and adds safer handling when `--sitename` differs from an existing project .env

## Changes
- switched env write format from legacy SITES to SITES_RULE (Traefik v3 style)
- added backward-compatible read fallback for legacy SITES
- added deprecation warning when legacy SITES is detected
- added `--confirm-site-mismatch` option for build, deploy, and upgrade
- when .env sites differ from CLI `--sitename`:
    - warns and aborts by default
    - proceeds only if --confirm-site-mismatch is provided
-    updated `README.md` for new flag and migration behavior

PR #1724, opened 2026-06-24
fix(easy-install): various issues

barredterra:
## first issue

Resolved through https://github.com/frappe/bench/pull/1701

## second issue

Upstream `frappe_docker` changed the custom image `Containerfile` to read `apps.json` via a BuildKit secret. Our script still only passed the old `APPS_JSON_BASE64` build arg, so the image was built with Frappe only, and `--install-app erpnext` failed with `No module named 'erpnext'`.

I updated the Docker build command to pass:

```sh
--secret id=apps_json,src=/absolute/path/to/apps.json
```

## third issue

`easy-install.py` defaulted `--frappe-branch` to `version-15`, while upstream `frappe_docker` now builds ERPNext from `version-16`. That produced the Frappe/ERPNext dependency mismatch. The later `pull access denied for custom-apps` happened because the script swallowed the Docker build failure and continued into deploy.

Changes made:

- Default `--frappe-branch` is now `version-16`.
- Bumped default python version from 3.11 to 3.14
- Bumped default node version from 18 to 24
- Docker image build/push failures now `sys.exit(1)` instead of continuing.

### DIFF
--- a/easy-install.py
+++ b/easy-install.py
@@ -86,6 +86,7 @@ def write_to_env(
     custom_tag: str = None,
 ) -> None:
     quoted_sites = ",".join([f"`{site}`" for site in sites]).strip(",")
+    sites_rule = " || ".join([f"Host(`{site}`)" for site in sites])
     example_env = get_from_env(frappe_docker_dir, "example.env")
     erpnext_version = erpnext_version or example_env["ERPNEXT_VERSION"]
     env_file_lines = [
@@ -100,6 +101,7 @@ def write_to_env(
         f"LETSENCRYPT_EMAIL={email}\n",
         f"SITE_ADMIN_PASS={admin_pass}\n",
         f"SITES={quoted_sites}\n",
+        f"SITES_RULE={sites_rule}\n",
         "PULL_POLICY=missing\n",
         f'BACKUP_CRONSTRING="{cronstring}"\n',
     ]
--- a/easy-install.py
+++ b/easy-install.py
@@ -694,7 +694,9 @@ def build_image(
         tags = ["custom-apps:latest"]
 
     apps_json_base64 = None
+    apps_json_path_abs = None
     try:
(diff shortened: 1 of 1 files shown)


revant:
@Rocket-Quack @DanielRadlAMR can you confirm?

easy-install.py
```diff
@@ -718,9 +722,15 @@ def build_image(
         f"--build-arg=PYTHON_VERSION={python_version}",
         f"--build-arg=NODE_VERSION={node_version}",
         f"--build-arg=APPS_JSON_BASE64={apps_json_base64}",
```
DanielRadlAMR: Why not just replace this line with the new `--secret` line?

DanielRadlAMR (changes requested):

DanielRadlAMR:
The fix for the third issue looks correct (it matches the approach in #1700, which addresses the same issue).

I left a review for the fix to the second issue. I'd recommend removing `APPS_JSON_BASE64` entirely, as keeping it around defeats the purpose of switching to build secrets in the first place.

Regarding the first issue, I'd wait for @Rocket-Quack. His #1701 already tackles the same issue.

Rocket-Quack:
Thx @DanielRadlAMR 
I think it's good that some progress has been made on this fix but personally I find the implementation a bit too limited.
Especially since there is no provision for handling existing environments like used old `SITES` Definition. 

IMHO #1701 is more suitable in this regard, but if the only way to implement the fix is through this PR then let’s go with it with the request to expand the documentation take legacy instances into account and also allow the old method of site naming.

easy-install.py
barredterra: Makes sense, dropped the old arg.

barredterra:
@Rocket-Quack happy to merge your PR, I'm sure it solves the problem better. I lack context and was just trying to get the CI to pass.

